PR #25, opened 2017-02-07
Initial changes towards getting builds for the construction of the Bears.

fermadeiral:
These changes are the first ones towards getting builds for the construction of the Bears.

First things first, I have changed the **ProjectScanner** class to also scan passing builds. The key changes consist in the "addition" of the ```getListOfBuildsFromProjectsByBuildStatus``` method, which is the previous ```getListOfFailingBuildFromProjects```, but now it receives the build status of interest. This build status is also used in the ```testBuild``` method, where the builds are actually tested for what we want---Simon wants failing builds, so the build status of interest is FAILED, while I want passing builds, so the build status of interest is PASSED.

I added the **ProjectInspector4Bears** class, which extends **ProjectInspector**. For now, **ProjectInspector4Bears** does basically the same thing than **ProjectInspector**, just ignoring Nopol in its code (I have rewritten the ```processRepair``` method in **ProjectInspector4Bears** to ignore Nopol). The reason I created this class is because I think I will create different steps to inspect the projects (for example, I don't need **NopolRepair** step, but I think I will need other types of steps).

In the **Launcher** class, I put the variables related to input args as global ones, and I added some conditions using **RepairMode.FORBEARS**, a new mode, since for Bears there are some things that are not necessary (e.g., repair by Nopol). I also added the ```inspectBuildsForBears``` method, which is similar to the ```cloneAndRepair``` method, but for the **ProjectInspector4Bears**.

In the **GoogleSpreadSheetFactory** class, I only changed the **SPREADSHEET_ID** and the **GOOGLE_SECRET_PATH** to test my changes. We have to define a way to integrate this.


repairnator/src/main/java/fr/inria/spirals/repairnator/serializer/gsheet/GoogleSpreadSheetFactory.java
```diff
@@ -33,8 +33,9 @@
     private Logger logger = LoggerFactory.getLogger(GoogleSpreadSheetFactory.class);
     private static final String APPLICATION_NAME = "RepairNator Bot";
     private static final File DATA_STORE_DIR = new File(System.getProperty("user.home"), ".credentials/sheets.googleapis.com-repairnator");
-    private static final String GOOGLE_SECRET_PATH = "./client_secret.json";
-    public static final String SPREADSHEET_ID = "1FUHOVx1Y3QZCAQpwcrnMzbpmMoWTUdNg0KBM3NVL_zA";
+    private static final String GOOGLE_SECRET_PATH = "/home/fernanda/git/librepair/repairnator/client_secret.json";
+    //public static final String SPREADSHEET_ID = "1FUHOVx1Y3QZCAQpwcrnMzbpmMoWTUdNg0KBM3NVL_zA";
+    public static final String SPREADSHEET_ID = "1MnRwoZGCxxbmkiswc0O6Rg43wJFTBc3bIyrNdTiBhQ4";
```
surli: Those lines should not be changed. We could use an argument for the `GOOGLE_SECRET_PATH` with a default value. And for `SPREADSHEET_ID` the constant is not used there but in the subclasses: we should either define them in subclasses or pass them as arguments.

repairnator/src/main/java/fr/inria/spirals/repairnator/Launcher.java
```diff
@@ -233,33 +245,40 @@ private void run(String[] args) throws JSAPException, IOException {
...
+        if (push || mode == RepairMode.FORBEARS) {
+            this.serializers.add(googleSpreadSheetInspectorSerializer);
         }
 
-        Launcher.LOGGER.debug("Start to scan projects in travis for failing builds...");
+        if (mode != RepairMode.FORBEARS) {
+        	Launcher.LOGGER.debug("Start to scan projects in travis for failing builds...");
+        } else {
+        	Launcher.LOGGER.debug("Start to scan projects in travis for passing builds...");
+        }
```
surli: This condition is unnecessary: we can use a global log message like "Start to scan projects in travis"

repairnator/src/main/java/fr/inria/spirals/repairnator/Launcher.java
```diff
@@ -282,6 +301,15 @@ private void mainProcess() throws IOException {
...
+                
+            case FORBEARS:
+            	buildList = scanner.getListOfPassingBuildsFromProjects(input);
+            	break;
+        }
+
+        if (mode == RepairMode.SLUG || mode == RepairMode.FORBEARS) {
+            GoogleSpreadSheetScannerSerializer scannerSerializer = new GoogleSpreadSheetScannerSerializer(scanner);
+            scannerSerializer.serialize();
         }
```
surli: Apparently there was a problem when solving conflict or you forget to remove the following lines

repairnator/src/main/java/fr/inria/spirals/repairnator/Launcher.java
```diff
@@ -336,6 +362,17 @@ private void mainProcess() throws IOException {
...
     }
+    
+    private void inspectBuildsForBears(List<Build> buildList, String workspace) throws IOException {
+    	initWorkspace(workspace);
+    	
+    	ProjectInspector4Bears inspector;
+    	for (Build build : buildList) {
+            inspector = new ProjectInspector4Bears(build, workspace, this.serializers, null, push, steps, mode);
+            inspector.setAutoclean(clean);
+            inspector.processRepair();
```
surli: We should change that method name: you don't want to repair, but run the inspector.

repairnator/src/main/java/fr/inria/spirals/repairnator/process/ProjectInspector.java
```diff
@@ -26,22 +26,22 @@
  * Created by urli on 26/12/2016.
  */
 public class ProjectInspector {
-    private final Logger logger = LoggerFactory.getLogger(ProjectInspector.class);
+    protected final Logger logger = LoggerFactory.getLogger(ProjectInspector.class);
```
surli: I don't agree: Loggers are private to their class you should not use inheritance of logger, but create your own in the subclass.

repairnator/src/main/java/fr/inria/spirals/repairnator/process/ProjectScanner.java
```diff
@@ -190,14 +195,14 @@ public int getTotalScannedBuilds() {
         return result;
     }
 
-    private boolean testBuild(Build build) {
+    private boolean testBuild(Build build, BuildStatus targetBuildStatus) {
         Repository repo = build.getRepository();
```
surli: In fact you're using the targetBuildStatus as a boolean value: so maybe the code will be more clear if you pass a boolean as an argument named "targetFailing"

surli:
The project_list should not be changed to contain only one project: you should create your own local_project_list. We should discuss about the ProjectInspector to see if you need to extend the class or to be able to specify the step pipeline more easily.

repairnator/src/main/java/fr/inria/spirals/repairnator/Launcher.java
fermadeiral: Yes, I agree. What about simple ```inspect``` or, like you said in your sentence, ```run```?

repairnator/src/main/java/fr/inria/spirals/repairnator/process/ProjectInspector.java
fermadeiral: Ok.

repairnator/src/main/java/fr/inria/spirals/repairnator/process/ProjectScanner.java
fermadeiral: Ok.

repairnator/src/main/java/fr/inria/spirals/repairnator/Launcher.java
surli: run seems good :)

repairnator/src/main/java/fr/inria/spirals/repairnator/serializer/gsheet/GoogleSpreadSheetFactory.java
fermadeiral: I don't follow your thought. First, it seems you want to do different things with ```GOOGLE_SECRET_PATH``` and ```SPREADSHEET_ID```, and then it seems you want to use the same solution for both. Another ideia, but I'm not sure if it's good: can we do something like a ```Constants``` class e depending on the ```mode```, these constants assume different values?

repairnator/src/main/java/fr/inria/spirals/repairnator/Launcher.java
fermadeiral: Ok.

repairnator/src/main/java/fr/inria/spirals/repairnator/serializer/gsheet/GoogleSpreadSheetFactory.java
surli: `GOOGLE_SECRET_PATH` is a path for a file, so it depends on your environment: we can define a default value but the real value should be pass as an argument when we launch the program. You should never have an absolute path in the code.

Concerning the `SPREADSHEET_ID` it's an open question: maybe we could use constants and get the right one depending on the mode, as you propose. Maybe you will create a specific class for the serializer and then each serializer has its own spreadsheet_id value.

repairnator/src/main/java/fr/inria/spirals/repairnator/Launcher.java
fermadeiral: I was checking this. You are right. The right "if" is the one that have two conditions.

fermadeiral:
Ok about the project_list.

Regarding the ```ProjectInspector```, maybe I don't need to extend such. I'm not sure how many steps I have to create. I'll think about it.

fermadeiral:
All points of the discussion were fixed on the "Reviews of the discussion on the first-pr" commit, except the point concerning ```SPREADSHEET_ID``` (I'm thinking about that).

PR #26, opened 2017-02-08
Allowing AbstractStep subclasses to get the launch mode from ProjectInspector instances.

fermadeiral:
Since some existing Repairnator steps will be useful for Bears, and some parts of them should have to behaviour differently for Bears, it's important to allow that the steps can check if they are running for Repairnator or Bears.

### DIFF
--- a/repairnator/src/main/java/fr/inria/spirals/repairnator/process/step/GatherTestInformation.java
+++ b/repairnator/src/main/java/fr/inria/spirals/repairnator/process/step/GatherTestInformation.java
@@ -153,24 +153,14 @@ public class GatherTestInformation extends AbstractStep {
         }
 
         if (this.getState() == ProjectState.HASTESTFAILURE) {
-        	if (!(inspector.getMode() == RepairMode.FORBEARS)) {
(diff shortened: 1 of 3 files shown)


repairnator/src/main/java/fr/inria/spirals/repairnator/process/step/GatherTestInformation.java
```diff
@@ -151,12 +153,24 @@ public FileVisitResult preVisitDirectory(Path file, BasicFileAttributes attrs) t
...
+        	} else {
+        		this.shouldStop = true;
+        	}
         } else {
-            this.shouldStop = true;
+        	if (!(inspector.getMode() == RepairMode.FORBEARS)) {
+        		this.shouldStop = true;
+        	} else {
+        		this.shouldStop = false;
+        	}
```
surli: You should simplify this very easily: you always take the contrary of my shouldStop status. 
Then add a line at the end like:
```java
this.shouldStop = (inspector.getMode() == RepairMode.FORBEARS) ? !this.shouldStop : this.shouldStop;
```

PR #35, opened 2017-02-10
Getting passing builds that fix bugs.

fermadeiral:
Getting passing builds that fix bugs by checking whether their previous builds are failing builds with failing tests. The state of these passing builds is `ProjectState.FIXERBUILD`.

Solver path used by Nopol is not required for FORBEARS mode anymore.


repairnator/src/main/java/fr/inria/spirals/repairnator/Launcher.java
```diff
@@ -243,7 +249,9 @@ private void run(String[] args) throws JSAPException, IOException {
             System.exit(-1);
         }
         this.checkToolsLoaded();
-        this.checkNopolSolverPath();
+        if (RepairMode.valueOf(arguments.getString("mode").toUpperCase()) != RepairMode.FORBEARS) {
```
surli: We can initialize the field mode right now and then you can use it in your condition

surli:
We should discuss about the design of this PR: you introduce creation of inspectors inside a step, so it introduces a new complexity I'd like to avoid.

repairnator/src/main/java/fr/inria/spirals/repairnator/Launcher.java
fermadeiral: Ok.

fermadeiral:
I'll try something and I'll send you. If the issue remains, then we discuss.

Issue #54, opened 2017-02-24
Refactoring: create an object to pass as input for the inspector of repairnator


surli:
Right now the scanner create a list of builds, but in the case of bears we want to get the current build, the previous one and the status (it is a case of failing/passing or passing/passing). 
Then we should have an object containing all information to pass to the inspector to create the proper pipeline

Issue #55, opened 2017-02-24
refactoring bear: filter projects in the passing/passing builds case


surli:
If two consequent builds are passing, we should filter them by getting the diff and looking if it concerns a test or not.

PR #58, opened 2017-02-24
Adding an object to pass as input for the inspector of repairnator (issue #54).

fermadeiral:


repairnator/src/main/java/fr/inria/spirals/repairnator/process/BuildToBeInspected.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Created by fernanda on 24/02/17.
+ */
+public class BuildToBeInspected {
+
+    private Build build;
+    private Build previousBuild;
+    // For RepairNator, the status is about the build, and for Bears it is about the previous build
+    private BuildStatus status;
```
surli: We should not use the build status but use another status, for example `ScannedBuildStatus` which would contain: BUILD_AND_FAIL, BUILD_WITH_TEST_CHANGES, ONLY_FAIL.

repairnator/src/main/java/fr/inria/spirals/repairnator/process/ProjectScanner.java
```diff
@@ -254,23 +235,39 @@ private boolean testBuild(Build build, boolean targetFailing) {
...
-        this.buildsId = new ArrayList<Integer>();
         for (Repository repo : repos) {
             List<Build> repoBuilds = BuildHelper.getBuildsFromRepositoryWithLimitDate(repo, this.limitDate);
 
             for (Build build : repoBuilds) {
                 this.totalScannedBuilds++;
                 if (testBuild(build, targetFailing)) {
-                    result.add(build);
-                    this.buildsId.add(build.getId());
+                    if (mode == RepairMode.FORBEARS) {
```
surli: why do you check the mode here?

repairnator/src/main/java/fr/inria/spirals/repairnator/process/ProjectScanner.java
```diff
@@ -297,19 +294,20 @@ public String readWorkspaceFromInput(String input) throws IOException {
...
+    public List<BuildToBeInspected> readBuildFromInput(String input) throws IOException {
+        List<BuildToBeInspected> buildsToBeInspected = new ArrayList<BuildToBeInspected>();
 
         Properties properties = getPropertiesFromInput(input);
         String buildId = properties.getProperty("buildid");
         if (buildId != null) {
             Build build = BuildHelper.getBuildFromId(Integer.parseInt(buildId), null);
             if (build != null) {
-                result.add(build);
+                BuildToBeInspected buildToBeInspected = new BuildToBeInspected(build, build.getBuildStatus());
```
surli: ok for that case for now, you could also add a value "unknown" to the status of the build (see my previous comment concerning status of buildtobeinspected)

repairnator/src/main/java/fr/inria/spirals/repairnator/process/BuildToBeInspected.java
fermadeiral: Ok.

repairnator/src/main/java/fr/inria/spirals/repairnator/process/ProjectScanner.java
fermadeiral: Because for RepairNator it is not necessary to do things with the previous build.

surli: oh ok I thought the else was closing testbuild() not this one. my bad.

repairnator/src/main/java/fr/inria/spirals/repairnator/process/ProjectScanner.java
fermadeiral: Ok.

fermadeiral:
I added the ScannedBuildStatus, but it is not used yet in the inspectors.

repairnator/src/main/java/fr/inria/spirals/repairnator/process/ProjectScanner.java
```diff
@@ -254,23 +235,43 @@ private boolean testBuild(Build build, boolean targetFailing) {
...
+                        // First of all, here it is checked if the current passing build has a
+                        // previous build---if doesn't, nothing is made as this build is not
+                        // useful for Bears
+                        Build previousBuild = BuildHelper.getLastBuildOfSameBranchOfStatusBeforeBuild(build, null);
+                        if (previousBuild != null) {
+                            this.logger.debug("Build: " + build.getId());
+                            this.logger.debug("Previous build: " + previousBuild.getId());
+
+                            if (previousBuild.getBuildStatus() == BuildStatus.FAILED) {
+                                BuildToBeInspected buildToBeInspected = new BuildToBeInspected(build, previousBuild, ScannedBuildStatus.ONLY_FAIL);
```
surli: Don't agree: it's not "ONLY FAIL" here: you just test the "current" build and its previous one, so it's another status like "PASSING_AND_FAIL" or so.

repairnator/src/main/java/fr/inria/spirals/repairnator/process/ProjectScanner.java
```diff
@@ -254,23 +235,43 @@ private boolean testBuild(Build build, boolean targetFailing) {
...
+                                buildsToBeInspected.add(buildToBeInspected);
+                            } else {
+                                if (previousBuild.getBuildStatus() == BuildStatus.PASSED) {
+                                    BuildToBeInspected buildToBeInspected = new BuildToBeInspected(build, previousBuild, ScannedBuildStatus.PASS);
```
surli: ok for now if you want but the objective is to remove this status type and to filter them to only get the "build with test changes" right?

repairnator/src/main/java/fr/inria/spirals/repairnator/process/inspectors/ProjectInspector.java
```diff
@@ -51,16 +52,15 @@
...
     private boolean isReproducedAsError;
-    protected Build previousBuild;
     protected boolean previousBuildFlag;
 
-    public ProjectInspector(Build failingBuild, String workspace, List<AbstractDataSerializer> serializers,
-            String nopolSolverPath, boolean push, RepairMode mode) {
-        this.build = failingBuild;
+    public ProjectInspector(BuildToBeInspected buildToBeInspected, String workspace, List<AbstractDataSerializer> serializers,
+                            String nopolSolverPath, boolean push, RepairMode mode) {
+        this.buildToBeInspected = buildToBeInspected;
```
surli: Actually you should also test in the run method that the buildToBeInspected status is the expected one

repairnator/src/main/java/fr/inria/spirals/repairnator/process/inspectors/ProjectInspector4Bears.java
```diff
@@ -1,124 +1,83 @@
...
+        // current passing build to ensure it is reproducible
+        AbstractStep cloneRepo = new CloneRepository(this);
+        AbstractStep buildRepo = new BuildProject(this);
+        AbstractStep testProject = new TestProject(this);
+        this.testInformations = new GatherTestInformation(this, new BuildShouldPass());
+
+        AbstractStep checkoutPreviousBuild = new CheckoutPreviousBuild(this);
+        AbstractStep buildRepoForPreviousBuild = new BuildProject(this);
+        AbstractStep testProjectForPreviousBuild = new TestProject(this);
+        if (this.getPreviousBuild().getBuildStatus() == BuildStatus.FAILED) {
```
surli: you should change this condition to use information from the inspectedBuild status.

repairnator/src/main/java/fr/inria/spirals/repairnator/process/ProjectScanner.java
fermadeiral: Ok.

repairnator/src/main/java/fr/inria/spirals/repairnator/process/ProjectScanner.java
fermadeiral: Yes, it's right. The thing is I'm trying to solve this current issue (issue #54). This thing you are talking about is the issue #55 .

repairnator/src/main/java/fr/inria/spirals/repairnator/process/inspectors/ProjectInspector.java
fermadeiral: Ok. For RepairNator inspector, I tested if the status of the buildToBeInspected is `ScannedBuildStatus.ONLY_FAIL`, which is the interesting status for RepairNator, but there is a moment I assigned the status `ScannedBuildStatus.UNKNOWN` when RepairNator is running (method `readBuildFromInput`, class `ProjectScanner`). I don't know if the builds with `ScannedBuildStatus.UNKNOWN` status is also interesting for you to be inspect. Is it?

repairnator/src/main/java/fr/inria/spirals/repairnator/process/ProjectScanner.java
surli: ok then, let see that in another PR.

repairnator/src/main/java/fr/inria/spirals/repairnator/process/ProjectScanner.java
```diff
@@ -297,19 +298,20 @@ public String readWorkspaceFromInput(String input) throws IOException {
...
-    public List<Build> readBuildFromInput(String input) throws IOException {
-        List<Build> result = new ArrayList<Build>();
+    public List<BuildToBeInspected> readBuildFromInput(String input) throws IOException {
+        List<BuildToBeInspected> buildsToBeInspected = new ArrayList<BuildToBeInspected>();
 
         Properties properties = getPropertiesFromInput(input);
         String buildId = properties.getProperty("buildid");
         if (buildId != null) {
             Build build = BuildHelper.getBuildFromId(Integer.parseInt(buildId), null);
             if (build != null) {
```
surli: ok in fact let's change there the condition with a `&& build.getStatus() == BuildStatus.FAIL`, then you can change the scannedbuildstatus to ONLY_FAIL too. And it should be ok with the inspector. WDYT?

repairnator/src/main/java/fr/inria/spirals/repairnator/process/inspectors/ProjectInspector.java
surli: Ok in fact I proposed a change in the scanner then to avoid this problem.

repairnator/src/main/java/fr/inria/spirals/repairnator/process/inspectors/ProjectInspector4Bears.java
fermadeiral: Ok, but the change here is temporary because I can't do that in the right way right now before solving the issue #55.

repairnator/src/main/java/fr/inria/spirals/repairnator/process/ProjectScanner.java
fermadeiral: I think it can work well.

repairnator/src/main/java/fr/inria/spirals/repairnator/process/inspectors/ProjectInspector4Bears.java
surli: Hmmm in the case of passing/passing ok, but here it's the specific "passing/fail" so it could be tested with the PASSING_AND_FAIL status you should add ^^

PR #59, opened 2017-02-27
Serializing process duration information in the FORBEARS mode (issue #57)

fermadeiral:


repairnator/src/main/java/fr/inria/spirals/repairnator/Launcher.java
```diff
@@ -373,7 +380,22 @@ private void mainProcess() throws IOException {
                     this.endProcessSerializer.serialize();
                 }
             } else {
-                inspectBuildsForBears(buildsToBeInspected, completeWorkspace);
+                List<ProjectInspector4Bears> inspectors = inspectBuildsForBears(buildsToBeInspected, completeWorkspace);
+
+                for (ProjectInspector4Bears inspector : inspectors) {
+                    if (inspector.isFixerBuildCase1()) {
+                        this.nbFixerBuildCase1++;
```
surli: why do you create a class field: you need those values only here, no?

repairnator/src/main/java/fr/inria/spirals/repairnator/process/step/BuildShouldFail.java
```diff
@@ -20,11 +21,17 @@ public boolean shouldBeStopped(GatherTestInformation gatherTestInformation) {
                     // its previous build is a failing build with failing tests
                     // and it can also be reproduced
                     gatherTestInformation.setState(ProjectState.FIXERBUILD_CASE1);
+                    if (inspector instanceof ProjectInspector4Bears) {
```
surli: I really don't like that, but I don't have an easy solution yet. So if you have ideas to avoid this cast...

repairnator/src/main/java/fr/inria/spirals/repairnator/Launcher.java
fermadeiral: Yes. I will fix that.

PR #63, opened 2017-03-01
Moving methods from CloneRepository class into the new GitHelper class.

fermadeiral:
See what you think about this. I'll extract a method from `CloneRepository` class to put in this new `GitHelper` class too.


repairnator/src/main/java/fr/inria/spirals/repairnator/process/step/AbstractStep.java
```diff
@@ -94,7 +94,7 @@ public void setState(ProjectState state) {
         }
     }
 
-    protected Logger getLogger() {
+    public Logger getLogger() {
```
surli: This logger should be only available in the current package (subclasses) not in other classes, so you shouldn't change the visibility here.

repairnator/src/main/java/fr/inria/spirals/repairnator/helpers/GitHelper.java
```diff
@@ -0,0 +1,190 @@
...
+            // get from github a patch between that commit and the targeted
+            // commit
+            // note that this patch could contain changes of multiple commits
+            GHCompare compare = ghRepo.getCompare(lastKnowParent, oldCommitSha);
+
+            showGitHubRateInformation(gh, step);
+
+            URL patchUrl = compare.getPatchUrl();
+
+            step.getLogger().debug("Retrieve commit patch from the following URL: " + patchUrl);
```
surli: You should use a logger of the current class, and display in the message the step name.

repairnator/src/main/java/fr/inria/spirals/repairnator/helpers/GitHelper.java
```diff
@@ -0,0 +1,190 @@
+package fr.inria.spirals.repairnator.helpers;
+
```
surli: I'm not fond of the package name: we already have some helpers for maven in repairnator/process/maven: we should follow the same pattern and put it in repairnator/process/git

repairnator/src/main/java/fr/inria/spirals/repairnator/process/step/CloneRepository.java
```diff
@@ -244,8 +64,8 @@ protected void businessExecute() {
 
                 git.fetch().setRemote("PR").call();
 
-                String commitHeadSha = testCommitExistence(git, prInformation.getHead().getSha());
-                String commitBaseSha = testCommitExistence(git, prInformation.getBase().getSha());
+                String commitHeadSha = GitHelper.testCommitExistence(git, prInformation.getHead().getSha(), this, build);
+                String commitBaseSha = GitHelper.testCommitExistence(git, prInformation.getBase().getSha(), this, build);
```
surli: You need to also replace it in the inherited steps (Checkout*) no?

repairnator/src/main/java/fr/inria/spirals/repairnator/process/step/CheckoutSourceCodeForPreviousBuild.java
```diff
@@ -23,14 +25,13 @@ protected void businessExecute() {
...
+            String commitCheckout = build.getCommit().getSha();
+            commitCheckout = GitHelper.testCommitExistence(git, commitCheckout, this, build);
             if (commitCheckout != null) {
-                this.getLogger()
-                        .debug("Get the commit " + commitCheckout + " for repo " + this.inspector.getRepoSlug());
+                this.getLogger().debug("Get the commit " + commitCheckout + " for repo " + this.inspector.getRepoSlug());
                 git.checkout().setName(commitCheckout).call();
-                git.checkout().setStartPoint(inspector.getPreviousBuild().getCommit().getSha()).addPath("src/main/java")
-                        .call();
+                git.checkout().setStartPoint(inspector.getPreviousBuild().getCommit().getSha()).addPath("src/main/java").call();
```
surli: Before the checkout you should also test the commit existence here.

repairnator/src/main/java/fr/inria/spirals/repairnator/process/step/CloneRepository.java
```diff
@@ -53,44 +48,12 @@ protected void businessExecute() {
...
-                String commitBaseSha = GitHelper.testCommitExistence(git, prInformation.getBase().getSha(), this, build);
-
-                if (commitHeadSha == null) {
-                    this.addStepError("Commit head ref cannot be retrieved in the repository: "
-                            + prInformation.getHead().getSha() + ". Operation aborted.");
-                    this.getLogger().debug(prInformation.getHead().toString());
+                boolean successfulMerge = GitHelper.mergeTwoCommitsForPR(git, build, prInformation, repository, this);
+                if (!successfulMerge) {
                     this.shouldStop = true;
-                    return;
```
surli: You should log there, that there is a problem with the merge

PR #69, opened 2017-03-02
Extracting CheckoutRepository step class from CloneRepository step class towards fixing the bug related to issue #65.

fermadeiral:


repairnator/src/main/java/fr/inria/spirals/repairnator/process/git/GitHelper.java
```diff
@@ -11,6 +11,7 @@
 import okhttp3.Response;
 import org.eclipse.jgit.api.Git;
 import org.eclipse.jgit.api.MergeCommand;
+import org.eclipse.jgit.api.MergeResult;
```
surli: do you need it?

repairnator/src/main/java/fr/inria/spirals/repairnator/process/step/CheckoutRepository.java
```diff
@@ -0,0 +1,82 @@
...
+import org.eclipse.jgit.api.Git;
+import org.eclipse.jgit.api.errors.GitAPIException;
+
+import java.io.File;
+import java.io.IOException;
+
+/**
+ * Created by fernanda on 02/03/17.
+ */
+public class CheckoutRepository extends AbstractStep {
```
surli: ok so this one should be abstract then: you don't want to execute the step as it.

repairnator/src/main/java/fr/inria/spirals/repairnator/serializer/AbstractDataSerializer.java
```diff
@@ -60,6 +60,12 @@ protected String getPrettyPrintState(ProjectState state, GatherTestInformation g
             case PREVIOUSVERSIONISNOTINTERESTING:
                 return "previous build is not interesting";
 
+            case BUILDCHECKEDOUT:
```
surli: I'm not sure this is the only place you have to add new cases... did you check on google sheets serializer?

repairnator/src/main/java/fr/inria/spirals/repairnator/process/git/GitHelper.java
fermadeiral: It was not on purpose. I was trying to do something after merge but then I gave up for now and I forgot to remove that import.

repairnator/src/main/java/fr/inria/spirals/repairnator/process/step/CheckoutRepository.java
fermadeiral: Ok.

repairnator/src/main/java/fr/inria/spirals/repairnator/serializer/AbstractDataSerializer.java
fermadeiral: Yes, this is not the only place.

Issue #67, opened 2017-03-01
Refactoring: allow to pass spreadsheet ID to repairnator command line


surli:
We should pass different command line argument through a config.ini file

fermadeiral:
Simon, please, edit this issue as you told me you want now to pass some arguments to RepairNator in a different way (file?) and then I can know what you want exactly.

surli:
I'm taking this one.

surli:
Done in #165

PR #76, opened 2017-03-03
Allowing to pass spreadsheet ID to repairnator command line (issue #67).

fermadeiral:
Actually, I put it as a required argument, because if I put one of the IDs as default, and if one of us forget of passing the spreadsheet ID by command line, we will mess up the spreadsheet of each other.


surli:
I don't want to add yet another argument to the command line. I will edit the issue to be more specific.

PR #87, opened 2017-03-06
Towards solving the issues #73 and #84

fermadeiral:
Main changes:
- Reorganization of `ProjectScanner` class;
- Addition of `thereIsDiffOnJavaSourceCode` method to check if there was at least one .java file changed between the build and its previous one (for issue #73);
- Division of `FORBEARS` mode into two modes (`SLUGFORBEARS` and `BUILDFORBEARS`) (for issue #84)---by the way, I think we have to think in other solution here;
- Rename `RepairMode` class to `LauncherMode`.

### DIFF
--- a/repairnator/repairnator-core/src/main/java/fr/inria/spirals/repairnator/LauncherMode.java
+++ b/repairnator/repairnator-core/src/main/java/fr/inria/spirals/repairnator/LauncherMode.java
@@ -3,6 +3,6 @@ package fr.inria.spirals.repairnator;
 /**
  * Created by urli on 25/01/2017.
  */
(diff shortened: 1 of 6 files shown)


surli:
> by the way, I think we have to think in other solution here

Agreed, in fact I think we should have a LauncherMode (Repair or Bear) and a FileMode (Slug or Build). I merge and a open a new issue for that.

fermadeiral:
> Agreed, in fact I think we should have a LauncherMode (Repair or Bear) and a FileMode (Slug or Build). I merge and a open a new issue for that.

D'accord.

Issue #90, opened 2017-03-06
Refactoring: split LauncherMode into LauncherMode and FileMode


surli:
We should have LauncherMode (Repair & Bear) and FileMode (Slug & Build)

PR #101, opened 2017-03-07
Splitting LauncherMode into LauncherMode and FileMode (issue #90)

fermadeiral:
I also reorganized the input arguments to follow the same order in the code in the different things (as declaration, initialization, etc) and other small things like that in the `Launcher` class.

### DIFF
--- a/repairnator/repairnator-core/src/main/java/fr/inria/spirals/repairnator/LauncherMode.java
+++ b/repairnator/repairnator-core/src/main/java/fr/inria/spirals/repairnator/LauncherMode.java
@@ -4,5 +4,5 @@ package fr.inria.spirals.repairnator;
  * Created by urli on 25/01/2017.
  */
 public enum LauncherMode {
-    BUILD, SLUG, SLUGFORBEARS, BUILDFORBEARS
+     REPAIRNATOR, BEARS
 }
--- a/repairnator/repairnator-core/src/main/java/fr/inria/spirals/repairnator/LauncherMode.java
+++ b/repairnator/repairnator-core/src/main/java/fr/inria/spirals/repairnator/LauncherMode.java
@@ -4,5 +4,5 @@ package fr.inria.spirals.repairnator;
  * Created by urli on 25/01/2017.
  */
 public enum LauncherMode {
-     REPAIRNATOR, BEARS
+     REPAIR, BEARS
 }
--- a/repairnator/repairnator-scanner/src/main/java/fr/inria/spirals/repairnator/Launcher.java
+++ b/repairnator/repairnator-scanner/src/main/java/fr/inria/spirals/repairnator/Launcher.java
@@ -47,19 +47,17 @@ public class Launcher {
     private JSAPResult arguments;
     private List<AbstractDataSerializer> serializers;
 
-    private String input;
-    private String workspace;
-    private int lookupDays;
-    private String output;
     private boolean debug;
-    private LauncherMode mode;
-    private boolean clean;
+    private String input;
+    private LauncherMode launcherMode;
+    private FileMode fileMode;
+    private String output;
+    private String workspace;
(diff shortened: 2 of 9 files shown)


repairnator/repairnator-core/src/main/java/fr/inria/spirals/repairnator/LauncherMode.java
```diff
@@ -4,5 +4,5 @@
  * Created by urli on 25/01/2017.
  */
 public enum LauncherMode {
-    BUILD, SLUG, SLUGFORBEARS, BUILDFORBEARS
+     REPAIRNATOR, BEARS
```
surli: I prefer REPAIR than REPAIRNATOR.

repairnator/repairnator-scanner/src/main/java/fr/inria/spirals/repairnator/scanner/ProjectScanner.java
```diff
@@ -99,15 +99,15 @@ public Date getLimitDate() {
         return limitDate;
     }
 
-    public List<BuildToBeInspected> getListOfFailingBuildsFromGivenBuildIds(String path, LauncherMode mode) throws IOException {
-        return getListOfBuildsFromGivenBuildIds(path, mode, true);
+    public List<BuildToBeInspected> getListOfFailingBuildsFromGivenBuildIds(String path, LauncherMode launcherMode) throws IOException {
```
surli: In fact, instead of passing the mode everywhere in the methods: wouldn't it easier to create a field for the mode and pass it in the constructor?

repairnator/repairnator-scanner/src/main/java/fr/inria/spirals/repairnator/Launcher.java
```diff
@@ -254,97 +264,88 @@ private void run(String[] args) throws JSAPException, IOException {
...
-            case SLUGFORBEARS:
-                buildsToBeInspected = scanner.getListOfPassingBuildsFromProjects(input, mode);
-                break;
-
-            case BUILDFORBEARS:
-                buildsToBeInspected = scanner.getListOfPassingBuildsFromGivenBuildIds(input, mode);
-                break;
+        if (this.launcherMode == LauncherMode.REPAIRNATOR) {
+            if (this.fileMode == FileMode.SLUG) {
+                buildsToBeInspected = scanner.getListOfFailingBuildsFromProjects(this.input, this.launcherMode);
```
surli: Actually, if we pass the launcherMode at construction time to the scanner, it would be able to select directly the right method based on the mode: it should be its responsibility, not the Launcher's one.

repairnator/repairnator-core/src/main/java/fr/inria/spirals/repairnator/LauncherMode.java
fermadeiral: Ok.

repairnator/repairnator-scanner/src/main/java/fr/inria/spirals/repairnator/scanner/ProjectScanner.java
fermadeiral: ;-)

repairnator/repairnator-scanner/src/main/java/fr/inria/spirals/repairnator/Launcher.java
fermadeiral: ;-) ;-)

fermadeiral:
> If you can, also update the documentation (the usage of Repairnator in repairnator/Readme.md)

I saw this just now. Wait, then.

fermadeiral:
I changed the order of some arguments in the launch_repairnator.sh file to match with other places.

Issue #91, opened 2017-03-06
"Reorganization" of the ProjectState possible values and their usage in RepairNator (for example, how we display them in the spreadsheets).


fermadeiral:

surli:
Fix with #134

PR #134, opened 2017-03-20
Reorganizing the ProjectState possible values and their usage in RepairNator (issue #91)

fermadeiral:


fermadeiral:
I also
- put the `CheckoutRepository` and `CheckoutType` in a new package (`fr.inria.spirals.repairnator.process.step.checkoutrepository`);
- added the "real state" in the Bears spreadsheet;
- changed some messages of the logger;
- and did other minor changes.

fermadeiral:
Some tests fail because of the reorganization of the ProjectState possible values:

```
Failed tests: 
  TestComputeSourceDir.testComputeSourceDir:62 
Expected: is <SOURCEDIRCOMPUTED>
     but: was <NOTTESTABLE>
  TestNopolRepair.testNopolRepair:82 
Expected: is <PATCHED>
     but: was <CLASSPATHNOTCOMPUTED>
  TestTestProject.testTestProjectWhenFailing:57 
Expected: is <false>
     but: was <true>

Tests run: 23, Failures: 3, Errors: 0, Skipped: 0
```

surli:
> Some tests fail because of the reorganization of the ProjectState possible values

Update the tests then.

fermadeiral:
> Update the tests then.

But did you review and agree with the new organization? I won't change the tests before that. By the way, I was thinking about it and I think that the problem may be not because of the reorganization. It doesn't make sense if it was.

PR #143, opened 2017-03-21
Change the run of Travis for JTravis

surli:


surli:
Apparently coveralls does not support when two different projects are called in the same run travis. So the figures only concerned jTravis here and not jTravis AND repairnator.

PR #144, opened 2017-03-21
Allowing Repairnator to scan projects in a given time interval (issue #140)

fermadeiral:


repairnator/repairnator-scanner/src/main/java/fr/inria/spirals/repairnator/scanner/ProjectScanner.java
```diff
@@ -35,15 +35,31 @@
...
+        this(launcherMode);
         this.lookupHours = lookupHours;
+        Calendar limitCal = Calendar.getInstance();
+        limitCal.add(Calendar.HOUR_OF_DAY, -this.lookupHours);
+        this.lookFromDate = limitCal.getTime();
+        this.lookToDate = new Date();
+    }
+
+    public ProjectScanner(Date lookFromDate, Date lookToDate, LauncherMode launcherMode) {
+        this(launcherMode);
```
surli: You should define completely this one (without calling this), and use it in the other one. The constructor with only `LauncherMode` should not be authorized as it will throw an exception if you try to get builds from it.

repairnator/repairnator-scanner/src/main/java/fr/inria/spirals/repairnator/scanner/Launcher.java
```diff
@@ -150,7 +169,14 @@ private void defineArgs() throws JSAPException {
...
         String googleSecretPath = this.arguments.getFile("googleSecretPath").getPath();
 
-        ProjectScanner scanner = new ProjectScanner(this.arguments.getInt("lookupHours"), launcherMode);
+        ProjectScanner scanner;
+        Date lookFromDate = this.arguments.getDate("lookFromDate");
+        Date lookToDate = this.arguments.getDate("lookToDate");
+        if (lookFromDate != null && lookToDate != null && lookFromDate.before(lookToDate)) {
+            scanner = new ProjectScanner(lookFromDate, lookToDate, launcherMode);
+        } else {
+            scanner = new ProjectScanner(this.arguments.getInt("lookupHours"), launcherMode);
```
surli: Actually we could even remove the constructor of Scanner with lookupHours and do the computing of the new date here. WDYT?

repairnator/repairnator-scanner/src/main/java/fr/inria/spirals/repairnator/scanner/ProjectScanner.java
fermadeiral: Oh, ok, I got it.

repairnator/repairnator-scanner/src/main/java/fr/inria/spirals/repairnator/scanner/Launcher.java
fermadeiral: I think it can work well.

PR #160, opened 2017-03-23
Refactor inspector and fix repo path

### LINKED ISSUE #154: Bug: if the pom.xml is not at the root of the directory it could lead to errors
There is a confusion between repo path and maven module path in the code. Then in `AbstractStep#testPomLocation` the inspector repoLocalPath could be changed leading to errors when doing git manipulation:

```
18:06:06.062 [main] ERROR f.i.s.r.p.step.SquashRepository - Error while committing (and/or pushing) new repairnator information. 
org.eclipse.jgit.errors.RepositoryNotFoundException: repository not found: /root/workspace/geoserver/geoserver/213800779/src
```

surli:
This PR intends to fix #154 and to refactor repairnator-pipeline in order to remove some stuff from `ProjectInspector`:
  - move some steps in their own package
  - create a JobStatus class to deal with dynamic properties of the job 
  - create a field in JobStatus to manage the path of the main pom.xml file (fix #154)


surli:
@fermadeiral could you review this one?

fermadeiral:
> @fermadeiral could you review this one?

Ok ;-)

repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/step/gatherinfocontract/GatherTestInformation.java
```diff
@@ -1,7 +1,8 @@
-package fr.inria.spirals.repairnator.process.step;
+package fr.inria.spirals.repairnator.process.step.gatherinfocontract;
```
fermadeiral: I think that the name of the package should be changed, maybe to "testinformation".

repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/step/gatherinfocontract/GatherTestInformation.java
```diff
@@ -34,23 +35,16 @@
     private Set<FailureLocation> failureLocations;
     private Set<String> failureNames;
     private String failingModulePath;
+    private boolean skipSettingStatusInformation;
 
     private ContractForGatherTestInformation contract;
 
-    public GatherTestInformation(ProjectInspector inspector, ContractForGatherTestInformation contract) {
+    public GatherTestInformation(ProjectInspector inspector, ContractForGatherTestInformation contract, boolean skipSettingStatusInformation) {
```
fermadeiral: Based on the way you instantiate GatherTestInformation in the inspectors, it seems that there is a correlation between `skipSettingStatusInformation` and the contracts `BuildShouldPass` and `BuildShouldFail`. If that is true, maybe a javadoc would be nice here.

repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/step/gatherinfocontract/GatherTestInformation.java
surli: Ok to change but I'd go more for sthg like `gatherinfo`: it's the step so should be an action, not the result.

fermadeiral: Great.

repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/step/gatherinfocontract/GatherTestInformation.java
surli: In fact this one is really for you to avoid putting in JobStatus wrong information. So maybe it's enough to put it in the contract, I really don't know. Tell me :)

fermadeiral: I agreed with what you did: `skipSettingStatusInformation` must be false when `BuildShouldFail`, because this is the scenario both of us want to save information. I don't see any better way to implement it now, but what I meant is that we should put some javadoc in the code explaining this correlation.

surli: Ok, so I keep the boolean argument, in case this correlation is not true anymore and I explain its goal, right?

fermadeiral: I think it is ok for now.

PR #347, opened 2017-09-21
Refactor improve buildtool detection

vianneyPayelle:
Adding build log
Improve Maven build detection


jtravis/src/main/java/fr/inria/spirals/jtravis/parsers/GradleLogParser.java
```diff
@@ -24,19 +28,27 @@ public TestsInformation parseLog(TravisFold outOfFolds) {
...
 
-        return this.createTestInformation();
+        return this.globalResults;
+    }
+
+    @Override
+    public List<TestsInformation> parseDetailedLog(TravisFold outOfFold) {
+        this.globalResults = new TestsInformation();
+        this.detailedResults = new ArrayList<TestsInformation>();
+        this.detailedResults.clear();
```
surli: Why do you clear the detailedResults?

jtravis/src/main/java/fr/inria/spirals/jtravis/parsers/MavenLogParser.java
```diff
@@ -12,80 +14,114 @@
...
+        }
 
-            Matcher mvnTestHeadMatcher = mvnTestHeadPattern.matcher(s);
+        this.computePassingTests();
+        return this.globalResults;
+    }
+
+    @Override
+    public List<TestsInformation> parseDetailedLog(TravisFold outOfFold) {
+        boolean inTestBlock = false;
```
surli: Actually here the code seems pretty similar than for `parseLog`, and it seems useless to compute twice the logs if you want first the global results and then the detailed one. 
I propose you add a boolean somewhere to know if the logs had been parsed, and if it's the case you return the results. Then you should have a private method `parseLog()` which returns nothing but only do the parsing job, and the returning method will only test the boolean call the private method if needed, and return the results.

PR #349, opened 2017-09-23
Start to implement travis API V3. 

vianneyPayelle:
Implementation of getBuildFromId function for TRAVIS V3 API.
Adding functions to existing ones to compare results.

### DIFF
--- a/jtravis/src/main/java/fr/inria/spirals/jtravis/entities/Pagination.java
+++ b/jtravis/src/main/java/fr/inria/spirals/jtravis/entities/Pagination.java
@@ -1,0 +1,19 @@
+package fr.inria.spirals.jtravis.entities;
+
+public class Pagination {
+    private int limit;
+    private int offset;
+    private int count;
+    private boolean is_first;
+    private boolean is_last;
+
+    public boolean isIs_last() { return is_last; }
+
+    public int getLimit() {
+        return limit;
+    }
+
+    public int getOffset() {
+        return offset;
+    }
+}
--- a/jtravis/src/main/java/fr/inria/spirals/jtravis/entities/Pagination.java
+++ b/jtravis/src/main/java/fr/inria/spirals/jtravis/entities/Pagination.java
@@ -4,10 +4,18 @@ public class Pagination {
     private int limit;
     private int offset;
     private int count;
+    private String href;
     private boolean is_first;
     private boolean is_last;
+    private Pagination next;
+    private Pagination prev;
+    private Pagination first;
+    private Pagination last;
+
 
     public boolean isIs_last() { return is_last; }
+
+    public boolean isIs_first() { return is_first; }
 
     public int getLimit() {
         return limit;
@@ -15,5 +23,25 @@ public class Pagination {
 
     public int getOffset() {
         return offset;
+    }
+
+    public String getHref() { return href; }
+
+    public void setHref(String href) { this.href = href; }
+
+    public Pagination getNext() {
+        return next;
+    }
+
+    public Pagination getFirst() {
+        return first;
+    }
+
+    public Pagination getLast() {
+        return last;
+    }
+
+    public Pagination getPrev() {
+        return prev;
     }
 }
--- a/jtravis/src/main/java/fr/inria/spirals/jtravis/helpers/AbstractHelper.java
+++ b/jtravis/src/main/java/fr/inria/spirals/jtravis/helpers/AbstractHelper.java
@@ -27,6 +27,7 @@ public abstract class AbstractHelper {
 
     private final static String USER_AGENT = "MyClient/1.0.0";
     private final static String ACCEPT_APP = "application/vnd.travis-ci.2+json";
+    private final static String TRAVIS_API_VERSION = "3";
 
     private String endpoint;
     private OkHttpClient client;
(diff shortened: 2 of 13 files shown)


jtravis/src/main/java/fr/inria/spirals/jtravis/helpers/AbstractHelper.java
```diff
@@ -109,6 +114,21 @@ protected String get(String url) throws IOException {
         return result;
     }
 
+    protected String getV3(String url) throws IOException {
+        Request request = this.requestBuilderV3(url).get().build();
```
surli: ok there is a misunderstanding here: I did not want to have access to the V2 and the V3 at the same time from the API. Then you should not duplicated the methods, but replace them. 
And even if I wanted to keep an access to V2, I don't want it like that: here the code is completely duplicated. If you need to keep both, add a boolean or an enum somewhere for qualifying the access to V3/V2 and use them throughout the calls.

jtravis/src/main/java/fr/inria/spirals/jtravis/helpers/BuildHelper.java
```diff
@@ -4,11 +4,13 @@
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
+import com.sun.org.apache.xpath.internal.operations.Bool;
```
surli: Why this import?

jtravis/src/main/java/fr/inria/spirals/jtravis/helpers/BuildHelper.java
```diff
@@ -52,42 +56,126 @@ protected static BuildHelper getInstance() {
     }
 
     public static Build getBuildFromId(int id, Repository parentRepo) {
-        String resourceUrl = getInstance().getEndpoint()+BUILD_ENDPOINT+id;
-
+        String resourceUrl ="";
+        if(Version.getVersionV3())
+        {
+            String ENDPOINTV3 = "build/";
```
surli: Why not creating a BUILD_ENPOINTV3 at the beginning of the class?

jtravis/src/main/java/fr/inria/spirals/jtravis/helpers/BuildHelper.java
```diff
@@ -52,42 +56,126 @@ protected static BuildHelper getInstance() {
...
         } catch (IOException e) {
             getInstance().getLogger().warn("Error when getting build id "+id+" : "+e.getMessage());
             return null;
         }
+
+    }
+
+    //@param offset set value to 0 for first request
+    private static void getBuildsFromSlugV3(String slug, List<Build> result, String QueryParameters, int offset, Date younger, Date older, int buildAfter, int buildBefore) {
+        slug = slug .replace("/","%2F");
```
surli: Use UrlHttpEncoded or equivalent

jtravis/src/main/java/fr/inria/spirals/jtravis/helpers/BuildHelper.java
```diff
@@ -52,42 +56,126 @@ protected static BuildHelper getInstance() {
...
+        String ENDPOINTV3 = "repo/"+slug+"/builds";
+        String resourceUrl = getInstance().getEndpoint()+ENDPOINTV3+"?offset="+offset;
+        Boolean byDate = false;
+        Boolean byNumber = false;
+        if(QueryParameters != "")
+            resourceUrl += "&"+QueryParameters;
```
surli: This param does not follow Java rules for naming parameters: you should not have a caps for starting

jtravis/src/main/java/fr/inria/spirals/jtravis/entities/Pagination.java
```diff
@@ -0,0 +1,47 @@
+package fr.inria.spirals.jtravis.entities;
+
+public class Pagination {
+    private int limit;
+    private int offset;
+    private int count;
+    private String href;
+    private boolean is_first;
+    private boolean is_last;
+    private Pagination next;
```
surli: That's kind of not true: those do only contains offset and limit. They don't have is_first/last next/last/first and so one. Create another class for them.

jtravis/src/main/java/fr/inria/spirals/jtravis/helpers/Version.java
```diff
@@ -0,0 +1,14 @@
...
+
+public class Version {
+
+    private static Boolean versionV3 = true;
+
+    public static void setVersion(Boolean version) {
+        versionV3 = version;
+    }
+
+    public static Boolean getVersionV3() {
```
surli: Should be called `isVersionV3`

jtravis/src/main/java/fr/inria/spirals/jtravis/helpers/Version.java
```diff
@@ -0,0 +1,14 @@
+package fr.inria.spirals.jtravis.helpers;
+
+public class Version {
+
+    private static Boolean versionV3 = true;
+
+    public static void setVersion(Boolean version) {
```
surli: Either you have a `setVersion` which can take an enum value, or you have `setV3` which takes a boolean value. I personnaly prefer the first one.

jtravis/src/main/java/fr/inria/spirals/jtravis/helpers/RepositoryHelper.java
```diff
@@ -30,31 +32,49 @@ protected static RepositoryHelper getInstance() {
     }
 
     public static Repository getRepositoryFromSlug(String slug) {
-        String resourceUrl = getInstance().getEndpoint()+REPO_ENDPOINT+slug;
-
+        String resourceUrl ="";
+        if(Version.getVersionV3()) {
+            String urlSlug = "";
+            try {
+                urlSlug = URLEncoder.encode(slug, "UTF-8");
```
surli: Basically you need to do it each time you use V3 and put a slug name, so extract this try/catch to an utility function.

PR #403, opened 2018-01-31
change the schema of naming librepair-experiments branches

surli:
branch name should not be identified by the patched buildid as we do not store it in the mongodb. Instead we should ALWAYS use the buggybuildid.

### DIFF
--- a/jtravis/src/test/java/fr/inria/spirals/jtravis/helpers/BuildHelperTest.java
+++ b/jtravis/src/test/java/fr/inria/spirals/jtravis/helpers/BuildHelperTest.java
@@ -249,7 +249,7 @@ public class BuildHelperTest {
     public void testGetTheLastBuildNumberOfADate() {
         Date date = TestUtils.getDate(2017, 3, 16, 22, 59, 59);
 
-        String slug = "Spirals-Team/librepair";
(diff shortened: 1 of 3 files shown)


surli:
@monperrus I need your advice on this one: the current behaviour is to name branches on librepair experiments using either:
- buggy build id + date of the bug + "bugonly" suffix, in case no human patch is found
- patched build id + date of the patch, in case a human patch is found

The problem is, as I don't store the patched build id in the DB, I cannot easily find the associated branch of a buggy build if a patch has been found.
That *should* not be a problem in the future as we also store the URL of the remote branch in the DB, but still I find having two different pattern for naming branch confusing.

WDYT? Should we rely on the old pattern naming or should we change it now for future branches?

monperrus:
Is it possible to to do a massive renaming of old branches? Or do we have to keep two different
naming conventions in the repo (old and new)?

surli:
> Is it possible to to do a massive renaming of old branches?

Yes it's possible. I could first merge the PR, start to use the new naming convention and then use a script to extract the buildid / date and rename from that.
We have to be absolutely sure about our naming format then.

monperrus:
per our discussion.

- we change the naming convention
- we don't rename anything

PR #412, opened 2018-02-20
Update repairnator readme and launcher scripts.

fermadeiral:


surli:
I don't agree with your changes there. 

1. I'm using almost everywhere the hierarchy with a /scripts folder containing all the scripts to launch repairnator

2. You cannot specify that REPAIRNATOR_GH_REPO_PATH will be github/repairnator because it will be, at least, github/repairnator/repairnator (there is a subfolder in this repo called repairnator) And most importantly I'm still using directory named github/librepair where I execute repairnator...

fermadeiral:
Ok. You can just close this pr then. I just can't run repairnator following the readme, but I already set up this to work for me, so...

PR #413, opened 2018-02-21
WiP: fix scanner related issues.

fermadeiral:


surli:
Could you explain what you're trying to fix there?

fermadeiral:
You mean in the entire pr or commit by commit?

surli:
In the entire PR.

fermadeiral:
It's not only one issue, and I fixed only two of them by now. When I finish, I will write an explanation summarizing all the fixes connected with the commits done. Is that ok for you? I will remove the WiP and I will request your review for you to know that I finished.

surli:
@fermadeiral I will very soon do a big refactoring of Repairnator to use the new version of JTravis I am working one (see: https://github.com/Spirals-Team/jtravis). So it would be really preferable for you to finish a first version of this pull request soon to avoid a big conflict with my own refactoring. 
And at least we can discuss the further change you want and I could add them directly in my own refactoring if you want.

surli:
Hmmmm. And apparently you're doing exactly what I was afraid of: you're mixing different concerns in this PR :-/ We should have discuss about it before. 

At the very minimum I see the following feature in the PR:
1. You want to be able to pass a couple of builds in argument to the dockerpool with a buggy build and its associate patch build
2. You want to be able to compute in a different manner the diff

I propose you to really split up this PR in two PRs. 
Especially since I don't agree with the first feature as it is proposed right now. I really want to simplify the usage of repairnator and adding more and more argument is not getting us in this direction. I suggest that we create a dedicating docker image for bears usage with the proper argument you need.

fermadeiral:
@surli ok.

PR #427, opened 2018-03-01
Fix date interval for scanning builds.

fermadeiral:
_How the scanner has worked when giving an interval of dates to scan builds:_

For both dates, i.e. `lookFrom` and `lookTo`, the scanner has got it at midnight. Thus, if I want to scan only one day, let's say 31/01/1991, I need to enter `lookFrom=31/01/1991` and `lookTo=01/02/1991`, which looks like strange for potential users (aka me). Moreover, despite the fact it is improbable to happen, if a build happened at 01/02/1991 00:00:00, I will get it, even that I wanted builds only from the day 31/01/1991. If I search two intervals, `lookFrom=31/01/1991` and `lookTo=01/02/1991`, and `lookFrom=01/02/1991` and `lookTo=02/02/1991`, in both searches, the scanner will return the same build that happened at 01/02/1991 00:00:00.

_Proposal:_

To make `lookTo` just before midnight. In the example, then, I would enter `lookFrom=31/01/1991` and `lookTo=31/01/1991`, and the scan would happen from 00:00:00 to 23:59:59 at 31/01/1991.

### DIFF
--- a/repairnator/repairnator-scanner/src/main/java/fr/inria/spirals/repairnator/scanner/Launcher.java
+++ b/repairnator/repairnator-scanner/src/main/java/fr/inria/spirals/repairnator/scanner/Launcher.java
@@ -224,14 +224,14 @@ public class Launcher {
         opt2.setShortFlag('f');
         opt2.setLongFlag("lookFromDate");
         opt2.setStringParser(dateStringParser);
-        opt2.setHelp("Specify the initial date to get builds");
+        opt2.setHelp("Specify the initial date to get builds (e.g. 01/01/2017). Note that the search starts from 00:00:00 of the specified date.");
         this.jsap.registerParameter(opt2);
 
         opt2 = new FlaggedOption("lookToDate");
         opt2.setShortFlag('t');
         opt2.setLongFlag("lookToDate");
         opt2.setStringParser(dateStringParser);
-        opt2.setHelp("Specify the final date to get builds");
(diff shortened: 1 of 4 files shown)


repairnator/repairnator-scanner/src/main/java/fr/inria/spirals/repairnator/scanner/Launcher.java
```diff
@@ -287,6 +287,15 @@ private void defineArgs() throws JSAPException {
...
         Date lookFromDate = this.arguments.getDate("lookFromDate");
         Date lookToDate = this.arguments.getDate("lookToDate");
+        if (lookToDate != null) {
+            Calendar calendar = Calendar.getInstance();
+            calendar.setTime(lookToDate);
+            calendar.set(Calendar.HOUR_OF_DAY, calendar.getMaximum(Calendar.HOUR_OF_DAY));
+            calendar.set(Calendar.MINUTE, calendar.getMaximum(Calendar.MINUTE));
+            calendar.set(Calendar.SECOND, calendar.getMaximum(Calendar.SECOND));
+            calendar.set(Calendar.MILLISECOND, calendar.getMaximum(Calendar.MILLISECOND));
+            lookToDate = calendar.getTime();
```
surli: Even if I trust you on this piece of code, I think it worthes it to encapsulate it in a function we can test easily :)

surli:
Ok with your proposal

fermadeiral:
@surli WDYT?

surli:
it looks good for me :) I'm waiting Travis to finish the build.

Issue #430, opened 2018-03-06
Question on notifiers.


fermadeiral:
For the PR #429, I'm currently working on notifiers. I realized that all launchers, except by the pipeline one, have the argument `notifyEndProcess`. However, I didn't find this working in the module `repairnator-realtime`, i.e. I didn't find the `notifyEnd` method from `EndProcessNotifier` being called.

Question on the module `repairnator-realtime`: what I just described is a bug or the notifier is not supposed to be in that module? My point is: either there is code missing (to make the notifier works) or there is more code than it should have (the argument `notifyEndProcess` and the other ones related to email notification shouldn't be pass to the launcher of that module).

surli:
Contrary to other modules of Repairnator such as the scanner or the dockerpool, `repairnator-realtime` aims at being always running without interruption.
Then it makes no sense to use a notifier for the end of the process. That's why I did not put it in.

fermadeiral:
Ok. Thus can I remove the arg `notifyEndProcess` from that module?

Another question: is the notifier engine being used somewhere in that module? The notifier engine has been initialized, but I still didn't find usage of this.

surli:
> Thus can I remove the arg notifyEndProcess from that module?

Yup. I've been a bit fast when doing my cut/paste there.

>  is the notifier engine being used somewhere in that module? The notifier engine has been initialized, but I still didn't find usage of this.

Right now, it's not used indeed. You can remove it too.

fermadeiral:
Deal.

PR #433, opened 2018-03-07
Remove useless stuff from repairnator realtime module (issue #430).

fermadeiral:
Changes due the discussion in issue #430.


surli:
Finally keep it as it is: Martin proposed to add a new argument for RTScanner in #431 and the notifyEndProcess will now makes sense.

Issue #443, opened 2018-03-12
Arg skip-failing in Scanner.


fermadeiral:
The arg `skip-failing` apparently is used only in Bears mode. I saw we use this in `ProjectScanner`, but apparently this arg is never really passed by the scripts. In fact, it seems to me we can make the things work in `ProjectScanner` without this arg. Do you remember why we put it? Do you see an actual usage for this? Because maybe we don't need it anymore, thus we could remove it.

surli:
I think we added it to only get the passing/passing builds from Bears when scanning. I really don't know if we might need it again: let me know...

fermadeiral:
Humm...
I will wait the PR #429 to be closed, then I will look to the code in the scanner and I will see if we can just remove it (by now, it seems to me that we can). Then I will open a new PR only for the changes related to this. I will keep this issue opened by now.

fermadeiral:
Closed by #456

fermadeiral:
Closed by #456

PR #456, opened 2018-03-16
Remove skip-failing arg from scanner

fermadeiral:
The arg `skip-failing` in scanner module is not necessary for Bears. In `ProjectScanner` there is already the variable `targetFailing` to make the things work properly for Bears.

This PR closes the issue #443.

### DIFF
--- a/repairnator/repairnator-scanner/src/main/java/fr/inria/spirals/repairnator/scanner/ProjectScanner.java
+++ b/repairnator/repairnator-scanner/src/main/java/fr/inria/spirals/repairnator/scanner/ProjectScanner.java
@@ -131,12 +131,7 @@ public class ProjectScanner {
     public List<BuildToBeInspected> getListOfBuildsToBeInspected(String path) throws IOException {
         this.scannerRunningBeginDate = new Date();
 
-        List<BuildToBeInspected> buildsToBeInspected;
-        if (RepairnatorConfig.getInstance().getLauncherMode() == LauncherMode.REPAIR) {
-            buildsToBeInspected = getListOfBuildsFromProjectsByBuildStatus(path, true);
-        } else {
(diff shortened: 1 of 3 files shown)


repairnator/repairnator-scanner/src/main/java/fr/inria/spirals/repairnator/scanner/ProjectScanner.java
```diff
@@ -270,11 +268,11 @@ private boolean testBuild(Build build, boolean targetFailing) {
                                 // testInfo can be null if the build tool is unknown
                                 if (testInfo != null && (testInfo.getFailing() > 0 || testInfo.getErrored() > 0)) {
                                     this.totalBuildInJavaFailingWithFailingTests++;
-                                    if (targetFailing) {
+                                    if (targetFailing) { // REPAIR mode
```
surli: Then we should completely refactor this to really use the RepairnatorConfig.LauncherMode here. It would be much more explicit about what we're doing.

fermadeiral: Ok. In fact I observed when doing the refactoring with `LauncherUtils` that the only module we don't use config is the scanner. It's probably a good idea to put it now. I will do the changes.

surli: > It's probably a good idea to put it now.

Completely!
I propose you do the changes in another PR: it will be easier to see the changes properly. Once merge you'll be able to change this one. Ok ?

fermadeiral: Ok.

fermadeiral:
@surli, I'm good with this PR, but I have a hard time to build with Travis because of the GitHub rate limit.

Issue #458, opened 2018-03-16
Align RepairnatorConfig with set_env_variable.sh and use config intensively in all modules.


fermadeiral:
`RepairnatorConfig` is highly related to the pipeline, but we use it for all modules now. What about creating one config for each module? Then we init the config for a given module with all args in that module, thus we don't need to retrieve specific args in any other location in the code?

surli:
The main purpose of RepairnatorConfig was actually to be used in every module... So I'm not sure about your proposal here.

fermadeiral:
Ok, let me try to elaborate...

Each module receives args. Some args are shared among modules, and some args are specific to single modules. For instance, in the realtime module, we have the arg `whitelist`. Observe that we use `RepairnatorConfig` field in that module, but we don't use it to store the value of the arg `whitelist`. Then, when we need this arg, we do like this: `this.arguments.contains("whitelist")`. As this example, it happens with a lot of args because in `RepairnatorConfig` we don't keep them. A simple approach would be to include all args of all modules in `RepairnatorConfig`, but I think it will be quite messy as things will be mixture. So what I was proposing would be to have in `RepairnatorConfig` only args that are shared among more than one or all modules, and to extend this `RepairnatorConfig` for each module to keep args specific to each module.

fermadeiral:
In a short and practical version, what I want is to be able to call the method `initConfig()` in each module and after that not to do things like this.arguments.contains("whitelist") ever again, because everything would be in config, so config would be intensively used.

surli:
Hmmmm. Actually I like the idea, but I maintain we should only keep one RepairnatorConfig: 

>  but I think it will be quite messy as things will be mixture

Not necessarily. Have a look on https://github.com/Spirals-Team/repairnator/blob/master/repairnator/scripts/set_env_variable.sh actually it should contain almost the same parameters.

fermadeiral:
Deal.

PR #461, opened 2018-03-19
Align RepairnatorConfig with modules' args and use config intensively in all modules

fermadeiral:
For issue #458.


fermadeiral:
@surli, I'm good with this PR.

Issue #446, opened 2018-03-13
Allow to set the list of projects to ignore


surli:
Currently `surli/failingProject` is ignored but hardcoded in the pipeline. It should be configurable.

surli:
Closed by #465

PR #465, opened 2018-03-20
Allow a configurable list of projects to be ignored for serialization in the pipeline.

fermadeiral:
This PR intends to close the issue #446. WDYT?

PS: I tested the changes with the scripts `launch_repairnator.sh` and `repair_buggy_build.sh`, and also directly by running the pipeline in IntelliJ. All these tests were done with the build 355425279, and I always got the new log message `[main] INFO  f.i.s.repairnator.pipeline.Launcher - The build 355425279 is from a project to be ignored (surli/failingproject), thus the pipeline deactivated serialization for that build.`. Nothing seems to be broken.


repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/pipeline/Launcher.java
```diff
@@ -243,6 +250,24 @@ private void initNotifiers() {
         this.notifiers.add(new FixerBuildNotifier(notifierEngines));
     }
 
+    private List<String> getListOfProjectsToIgnore() {
+        List<String> result = new ArrayList<>();
+        if (this.config.getProjectsToIgnoreFilePath() != null) {
+            File file = new File(this.config.getProjectsToIgnoreFilePath());
+            try {
+                BufferedReader reader = new BufferedReader(new FileReader(file));
```
surli: Actually this code can really be simplified by using `Files#readAllLines`: https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#readAllLines-java.nio.file.Path-

See an example of this usage here: https://github.com/Spirals-Team/repairnator/blob/4566d6dc73f96acc0de686acccdaf9e8b4c61c44/repairnator/repairnator-realtime/src/main/java/fr/inria/spirals/repairnator/realtime/RTScanner.java#L89

surli:
Great thanks!

fermadeiral:
@surli, see if it is better. I couldn't simplify that much because I need to put the projects' names to lower case =/

repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/pipeline/Launcher.java
```diff
@@ -243,6 +251,20 @@ private void initNotifiers() {
         this.notifiers.add(new FixerBuildNotifier(notifierEngines));
     }
 
+    private List<String> getListOfProjectsToIgnore() {
+        List<String> result = new ArrayList<>();
+        if (this.config.getProjectsToIgnoreFilePath() != null) {
+            try {
+                for (String line : Files.readAllLines(new File(this.config.getProjectsToIgnoreFilePath()).toPath())) {
```
surli: Please use a local variable for this `Files.readAllLines(new File(this.config.getProjectsToIgnoreFilePath()).toPath())`: it's more readable and can be useful in case of bug

repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/pipeline/Launcher.java
```diff
@@ -243,6 +251,20 @@ private void initNotifiers() {
...
+                    result.add(line.trim().toLowerCase());
+                }
+            } catch (IOException e) {
+                throw new RuntimeException("Error while reading projects to be ignored from file: " + this.config.getProjectsToIgnoreFilePath(), e);
```
surli: I don't agree: in this case we should just log an error and use an empty list: it's overkill to throw a runtime exception for an optional feature like this one.

repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/pipeline/Launcher.java
fermadeiral: Done,

repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/pipeline/Launcher.java
fermadeiral: Done.

repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/pipeline/Launcher.java
```diff
@@ -255,11 +255,13 @@ private void initNotifiers() {
...
             try {
-                for (String line : Files.readAllLines(new File(this.config.getProjectsToIgnoreFilePath()).toPath())) {
+                List<String> lines = Files.readAllLines(new File(this.config.getProjectsToIgnoreFilePath()).toPath());
+                for (String line : lines) {
                     result.add(line.trim().toLowerCase());
                 }
             } catch (IOException e) {
-                throw new RuntimeException("Error while reading projects to be ignored from file: " + this.config.getProjectsToIgnoreFilePath(), e);
+                LOGGER.error("Error while reading projects to be ignored from file "+this.config.getProjectsToIgnoreFilePath(), e);
+                result.clear();
```
surli: Hmmm actually we do not need to clear the list. I just meant that we should return the list.

fermadeiral: Done.

PR #484, opened 2018-03-28
Use JTravis V2

surli:


surli:
This PR will be stuck for some time: we cannot get right now the language from Travis API V3. I'll work on it later.

PR #487, opened 2018-04-04
Fix syntax error in launch_repairnator.sh

fermadeiral:
When running `launch_repairnator.sh`, I got "./launch_repairnator.sh: line 123: syntax error: unexpected end of file". This PR fixes that.


fermadeiral:
PS: The build failed but I do not understand why. I change nothing related to what failed. Was the test `TestRepairnatorConfig.testGithubOauth` already failing?

PR #506, opened 2018-04-09
add docker log command in repair_buggy_build.sh

monperrus:


surli:
Another way to do it is to directly call the docker logs command: it will block the script until it's done. Maybe I can add an argument to the script:

`./repair_buggy_build.sh <build-id>` -> display directly the docker logs
`./repair_buggy_build.sh -d <build-id>` -> daemon mode, only display the docker container id and the command to display logs

WDYT?

PR #508, opened 2018-04-10
Fix launch_scanner.sh

fermadeiral:


surli:
I actually fixed it ;)

PR #520, opened 2018-04-12
Bump all versions

surli:
Repairnator is no more compliant with first release: lots of arguments have changed and the scripts cannot be used anymore with the 1.0. So I proposed to bump everything to version 2.0 right now. So contrary to what I said, next release of the scanner would be 2.0. WDYT @fermadeiral ?


fermadeiral:
Can we merge this after being sure the scanner is ok? I'm asking this because I'm afraid the scanner is still not ok, and then even the scanner will not be ready in 2.0.

PR #526, opened 2018-05-02
bump npefix version

monperrus:


surli:
I released maven-repair. It does not work but here it's clearly a regression: another test with NPEFix is passing. This one is not passing because I'm testing it against a multi-module project. If I remember well maven-repair was not able to repair multi-module projects, right @tdurieux ? 

If you want to reproduce the error, I'm checking it with the following project: https://github.com/surli/failingProject/tree/multi-module-npe

tdurieux:
normally it should, it is the purpose of the tool :/

PR #533, opened 2018-05-09
Change the output and log folder names.

fermadeiral:
The issue this PR addresses was discussed by email. See what you think on the changes.

### DIFF
--- a/repairnator/scripts/utils/create_structure.sh
+++ b/repairnator/scripts/utils/create_structure.sh
@@ -11,10 +11,8 @@ mkdir -p $ROOT_LOG_DIR
 mkdir -p $ROOT_BIN_DIR
 mkdir -p $ROOT_OUT_DIR
 
-if [ ! -z "$RUN_ID" ]; then
-    export REPAIR_OUTPUT_PATH=$REPAIR_OUTPUT_PATH"_$RUN_ID"
-    export LOG_DIR=$LOG_DIR"_$RUN_ID"
-fi
+export REPAIR_OUTPUT_PATH=$REPAIR_OUTPUT_PATH"_$RUN_ID"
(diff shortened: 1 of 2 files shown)


repairnator/scripts/utils/create_structure.sh
```diff
@@ -11,7 +11,12 @@ mkdir -p $ROOT_LOG_DIR
 mkdir -p $ROOT_BIN_DIR
 mkdir -p $ROOT_OUT_DIR
 
-echo "Create output dir : $REPAIR_OUTPUT_PATH"
+if [ ! -z "$RUN_ID" ]; then
```
surli: I'm not sure about this check: the RUN_ID should always exist. However I'm not sure it's value is available in this script.

fermadeiral: Simon, I run the script `./launch_scanner.sh` and I got this:

Set environment variables
This will be run with the following RUN_ID: 0202696c-2c76-4eda-8a4d-032e38445450
Create output dir: /home/fernanda/repairnator/out/2018-05-09_081620_0202696c-2c76-4eda-8a4d-032e38445450
Create log directory: /home/fernanda/repairnator/logs/2018-05-09_081620_0202696c-2c76-4eda-8a4d-032e38445450
Copy jars
[...]

Thus, the value is available in that script.

About the check, I'm also not sure. I put it only to be sure that the `RUN_ID` was already set before to change the value of existing environment variables.

surli: ok for the value of RUN_ID. IMHO You can remove the check.

fermadeiral: Done.

PR #534, opened 2018-05-09
Log and print pair of builds for Bears

fermadeiral:


repairnator/repairnator-scanner/src/main/java/fr/inria/spirals/repairnator/scanner/Launcher.java
```diff
@@ -263,7 +270,11 @@ private void printToFile(List<BuildToBeInspected> listOfBuilds) {
 
     private void printToStdout(List<BuildToBeInspected> listOfBuilds) {
         for (BuildToBeInspected buildToBeInspected : listOfBuilds) {
-            System.out.println(buildToBeInspected.getBuggyBuild().getId());
+            if (this.config.getLauncherMode() == LauncherMode.REPAIR) {
+                System.out.println(buildToBeInspected.getBuggyBuild().getId());
+            } else {
+                System.out.println(buildToBeInspected.getBuggyBuild().getId() + Utils.COMMA + buildToBeInspected.getPatchedBuild().getId());
```
surli: We'll have the same error as before here: getId() return an int and comma is a char, so it will do the addition instead of concatenate the string

fermadeiral: Fixed.

PR #539, opened 2018-05-10
Fix LauncherUtils.checkEnvironmentVariable()

fermadeiral:
When running the pipeline module (directly in intelliJ), the execution crashes when it reaches the first step related to maven because `M2_HOME` was not set. Since this is an issue that should be captured at the beginning, I found a bug in the checks when the `Launcher`starts. This PR is supposed to fix that.

### DIFF
--- a/repairnator/repairnator-core/src/main/java/fr/inria/spirals/repairnator/LauncherUtils.java
+++ b/repairnator/repairnator-core/src/main/java/fr/inria/spirals/repairnator/LauncherUtils.java
@@ -284,11 +284,9 @@ public class LauncherUtils {
     }
 
     public static void checkEnvironmentVariable(String envVariable, JSAP jsap, LauncherType launcherType) {
-        if (launcherType != LauncherType.PIPELINE) {
-            if (System.getenv(envVariable) == null || System.getenv(envVariable).equals("")) {
-                System.err.println("You must set the following environment variable: "+envVariable);
-                LauncherUtils.printUsage(jsap, launcherType);
-            }
+        if (System.getenv(envVariable) == null || System.getenv(envVariable).equals("")) {
+            System.err.println("You must set the following environment variable: "+envVariable);
+            LauncherUtils.printUsage(jsap, launcherType);
         }
     }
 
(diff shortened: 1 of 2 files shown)


repairnator/repairnator-core/src/main/java/fr/inria/spirals/repairnator/LauncherUtils.java
```diff
@@ -284,11 +284,9 @@ public static void checkArguments(JSAP jsap, JSAPResult arguments, LauncherType
     }
 
     public static void checkEnvironmentVariable(String envVariable, JSAP jsap, LauncherType launcherType) {
-        if (launcherType != LauncherType.PIPELINE) {
-            if (System.getenv(envVariable) == null || System.getenv(envVariable).equals("")) {
-                System.err.println("You must set the following environment variable: "+envVariable);
-                LauncherUtils.printUsage(jsap, launcherType);
-            }
+        if (System.getenv(envVariable) == null || System.getenv(envVariable).equals("")) {
```
surli: This won't work: this method is used also to check GITHUB_OAUTH and the env variable is not available when running the pipeline through the docker image: we use the Github token through an argument (the idea behind it is to avoid keeping Github token as an env variable to avoid leaking it in case of env variable dumping in logs).

In short, to avoid any confusion here, we should have two specific methods: one to check Github Oauth and the other to check M2_HOME.

fermadeiral: Oh, ok. I will fix that.

fermadeiral: I tried a different approach without creating two separate methods. I liked this approach because `checkEnvironmentVariable` is only called from `checkArguments` now. See what you think. If you still prefer two separate methods, I will change that.

surli: I'm not a huge fan but it should work. Let's keep it for now.

PR #560, opened 2018-05-16
Refactor launch_repairnator.sh

fermadeiral:


repairnator/scripts/launch_dockerpool.sh
```diff
@@ -0,0 +1,82 @@
...
+    then
+        echo ""
+    else
+        echo "$1 $2 "
+    fi
+  fi
+}
+
+if [ "$#" -eq 0 ]; then
+    echo "A list of build ids must be provided"
```
surli: Put a usage at the beginning of the script. See what I did in `repair_buggy_build.sh`

repairnator/scripts/launch_dockerpool.sh
```diff
@@ -0,0 +1,82 @@
...
+NB_LINES=`wc -l < $REPAIRNATOR_BUILD_LIST`
+
+if [[ $NB_LINES == 0 ]]; then
+    echo "No build has been found. Stop the script now."
+    exit 0
+else
+    echo "$NB_LINES builds have been found."
+fi
+
+if [ "$#" -eq 1 ]; then
```
surli: What's the purpose of this if? If `$# > 1` you don't execute those lines? There's a problem here.

repairnator/scripts/launch_dockerpool.sh
```diff
@@ -0,0 +1,82 @@
...
+    if [ "$2" == "null" ];
+    then
+        echo ""
+    else
+        echo "$1 $2 "
+    fi
+  fi
+}
+
+if [ "$#" -eq 0 ]; then
```
surli: If there should be exactly one argument, maybe use `-ne 1` instead of `-eq 0`

repairnator/scripts/launch_scanner.sh
```diff
@@ -16,16 +16,18 @@ function ca {
   fi
 }
 
-SCRIPT_DIR="$(dirname "${BASH_SOURCE[0]}")"
+if [ "$#" -eq 0 ]; then
```
surli: I don't understand this if...

repairnator/scripts/launch_dockerpool.sh
fermadeiral: > If $# > 1 you don't execute those lines?

Yes. It's because the `launch_dockerpool.sh` (and also `launch_scanner.sh`) can be executed from `launch_repairnator.sh`. So, if this is the case, the first stuff to do, like to init and create the structure, should be done only once (i.e. in `launch_repairnator.sh`), thus those lines can't be executed. See in `launch_repairnator.sh` how I call `launch_dockerpool.sh` and `launch_scanner.sh`. There is an extra arg to control what I just described. I found no way better to do that. If you have any suggestion, please, let me know.

repairnator/scripts/launch_dockerpool.sh
fermadeiral: Actually `launch_dockerpool.sh` needs to receive one arg (the list of builds), but it can also receive one more arg if it's running throught `launch_repairnator.sh`.

repairnator/scripts/launch_scanner.sh
fermadeiral: `launch_scanner.sh` doesn't require any arg. But if this is running throught `launch_repairnator.sh`, it can't execute those lines because the things were already initialized.

repairnator/scripts/launch_dockerpool.sh
fermadeiral: I will do it.

repairnator/scripts/launch_dockerpool.sh
surli: In that case, 
  1. We should be calling only init_scripts.sh which should contain also content of create_structure.sh
  2. When init_script.sh is called, it should initialize a variable like "REPAIRNATOR_INITIALIZED" and you should test the value of this variable.

repairnator/scripts/launch_scanner.sh
surli: See my answer above. Current design is really unclear.

fermadeiral: I agree.

repairnator/scripts/launch_dockerpool.sh
fermadeiral: Perfect.

repairnator/scripts/launch_dockerpool.sh
surli: > it can also receive one more arg

Which one? I cannot see it in `launch_repairnator.sh`

fermadeiral: Now `launch_dockerpool.sh` receives only one arg, the list of builds. The other one was to control if the things had been initialized, but with your approach we don't need it anymore.

surli: So we can use `-ne 1` here...

fermadeiral: Done.

surli:
Have you tested the scripts before I merge?

fermadeiral:
> Have you tested the scripts before I merge?

I tested more the ones which were the center of the refactoring: `launch_repairnator.sh`, `launch_scanner.sh` and `launch_dockerpool.sh`. They seem to work properly. The others I just run until the main process starts (to see if the initialization is ok, which is what I changed on them). I had an issue with `check_branches.sh` but I just fixed it. They also seem to work properly.

surli:
OK don't bother too much with check_branches.sh in the future: it's really deprecated, it needs to be refactored.

PR #564, opened 2018-05-17
Fix pipeline docker logger level

fermadeiral:
I have not obtained debug log messages from the pipeline when running it from dockerpool. This PR proposes a solution for that. I changed only for Bears because I don't know if you are interested on this. If you are, and if you like the proposed changes, let me know, then I'll do the changes for Repair too.


repairnator/docker-images/bears-dockerimage/logback.xml
```diff
@@ -14,15 +14,15 @@
 	</appender>
 
 	<logger name="fr.inria.spirals.jtravis.helpers" level="info" />
-	<logger name="fr.inria.spirals.repairnator" level="info" />
+	<logger name="fr.inria.spirals.repairnator" level="${LOG_LEVEL:-info}" />
```
surli: I don't understand this notation. The ":-info" is a default value?

repairnator/docker-images/bears-dockerimage/logback.xml
```diff
@@ -14,15 +14,15 @@
 	</appender>
 
 	<logger name="fr.inria.spirals.jtravis.helpers" level="info" />
```
surli: It should be changed to `fr.inria.jtravis`

repairnator/docker-images/bears-dockerimage/logback.xml
fermadeiral: Yes.

repairnator/docker-images/bears-dockerimage/logback.xml
fermadeiral: Done.

PR #569, opened 2018-05-18
Extract PushProcessEnd step for pipeline

fermadeiral:
This PR adds the step `PushProcessEnd` in the pipeline module, which contains the code of the method `lastPush()` in `AbstractStep`. This step is supposed to be executed independently if the pipeline was entirely executed with success or not. For that reason, instead of adding this step in the actual pipeline construction, I added a field `finalStep` in `ProjectInspector` to keep this new step, which is executed when the pipeline needs to finish.
WDYT?


fermadeiral:
This PR is still a WIP because I will wait the PR #565 be closed first (there is a method there that can be used here to avoid the insertion of duplicate code).

surli:
I like the idea of having some specific steps to be run at the end of the pipeline.

fermadeiral:
> I like the idea of having some specific steps to be run at the end of the pipeline.

Good, me too. I think this type of change will make the code clear. This PR is probably just a start.

Now this PR is ready for review.

fermadeiral:
I extended the intention of this PR a little bit, but it's still related to the last push and the creation of the step `PushProcessEnd`.

I did two main changes:
- I put the code containing in the method `lastPush()` from `AbstractStep` in the new step `PushProcessEnd`. This new step is executed exactly when the method `lastPush()` was being executed, i.e. when the pipeline needs to finish. Well, I don't need to argue on that change because you had already put a comment in the code saying that the code in the method `lastPush()` shouldn't be where it was anyway, so let's move to the other change...
- Then, I "merged" the code from `PushIncriminatedBuild` into `PushProcessEnd`. Now I need to argue for you to understand what I thought when I did that: I think it was confused to push in different moments, i.e. the overall vision of the pushing feature in repairnator was difficult to follow (it still is, because of the commit steps, but this is for another PR ;-). It also complicates the checks regarding PushState when `PushProcessEnd` was added, and I also tried to avoid to create more PushStates. With the changes I did in this PR, it's clear when the new branch is created and when all commits are pushed: it's in the end of the pipeline, when `PushProcessEnd` is run.

WDYT?

Testing:
- All tests are passing
- I run `./repair_buggy_build.sh 352395977` with a image I created containing the changes of this PR: https://hub.docker.com/r/fermadeiral/repairnator/builds/bp4hzet2asbp9asuwjbopgc/
I got the branch https://github.com/fermadeiral/bears-collection/tree/surli-test-repairnator-352395977-20180312-163748
I thought a little bit strange the organization of the files in the repo, so I thought I had broken something with this PR.
Then I run `./repair_buggy_build.sh 352395977` again but with surli/repairnator:$PIPELINE_VERSION. I got the branch https://github.com/fermadeiral/bears-collection-fp/tree/surli-test-repairnator-352395977-20180312-163748, which looks like the same thing as the other one. Thus, if there is some bug, it's not this PR that introduced it. Could you check those branches and tell me if there is something wrong? Because if so, we need to fix that fast, but not in this PR I think.

surli:
The structure of the pushed branch is not "normal" but it was indeed already the case in the older version of repairnator so it's ok if you get the same. 
Please fix the conflict then I'll merge it.

fermadeiral:
Simon, please, wait the build to pass in https://github.com/fermadeiral/repairnator/pull/3 before merging.

fermadeiral:
BUILD PASSING ;-)

PR #570, opened 2018-05-18
Remove step from Bears pipeline

fermadeiral:
The removed step had been used to commit Nopol info, but we don't need it anymore.


surli:
And didn't you want to make the inversion between the two last steps?

PR #575, opened 2018-05-18
Fix broken build

### LINKED ISSUE #574: regression: TestInitRepoToPush.testInitRepoShouldRemoveNotificationInTravisYML:155
@fermadeiral your last PR broke a test. Could you have a look?

fermadeiral:
This PR is supposed to fix #574

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,3 +11,7 @@ script: "./.travis/travis-run.sh"
 
 after_success:
   - "./.travis/travis-deploy.sh"
+
+branches:
+  only:
+  - master
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,3 @@ script: "./.travis/travis-run.sh"
 
 after_success:
   - "./.travis/travis-deploy.sh"
-
-branches:
-  only:
-  - master
--- a/repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/git/GitHelper.java
+++ b/repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/git/GitHelper.java
@@ -39,7 +39,9 @@ import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.io.FileFilter;
+import java.io.FileWriter;
 import java.io.IOException;
+import java.nio.file.Files;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.HashSet;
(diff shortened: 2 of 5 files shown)


.travis.yml
```diff
@@ -11,7 +11,3 @@ script: "./.travis/travis-run.sh"
 
 after_success:
   - "./.travis/travis-deploy.sh"
-
-branches:
-  only:
-  - master
```
surli: Please remove this change.

fermadeiral:
The build passed on my side: https://travis-ci.org/fermadeiral/repairnator/builds/380683739

surli:
I don't agree with your solution. If I understand well, the problem is that we copy the entire directory each time before committing, but we only analyze once that we should change the notification part of `.travis.yml` right?

So in fact we should analyze each time the .travis.yml file.

.travis.yml
fermadeiral: Yeah, I didn't find another way to trigger build on different branches than master on my side, so I had to do that to build a build, but I just reverted it.

fermadeiral:
> I don't agree with your solution.

My solution is based on what was already in the code before the refactoring I did, which introduced the bug. At the end, if I haven't introduced the bug, this PR now plus the other one contains only a refactoring.

> So in fact we should analyze each time the .travis.yml file.

I don't know what to answer about that. Just tell me what to do for us to be able to make repairnator to pass.
WDYT about this: to move the method InitRepoToPush.removeNotificationFromTravisYML to `GitHelper`? Then, we would call this every time after the call to `gitHelper.copyDirectory()`, which would be hardcoded with the file extensions ".git" and ".m2".

surli:
> My solution is based on what was already in the code before the refactoring I did, which introduced the bug. At the end, if I haven't introduced the bug, this PR now plus the other one contains only a refactoring.

Yeah I know that, sorry I was a bit direct here, but it's still ugly to copy the whole repo and I don't like it. But right now I don't want to change it. Just to explain my reaction ;)

> WDYT about this: to move the method InitRepoToPush.removeNotificationFromTravisYML to GitHelper? Then, we would call this every time after the call to gitHelper.copyDirectory()

This is what I had in mind yes.

> which would be hardcoded with the file extensions ".git" and ".m2".

Actually you can keep your knew argument `fileExtension` just rename it to something like `excludedFilenames` or something like that. But yeah call it each time with .git and .m2 excluded.

fermadeiral:
@surli tell me what you think.

I'm not really able to build a build on my side without those changes I had done in the travis file. Do you know if there is another way for me to do that?

PR #584, opened 2018-05-24
Commit changed tests in a distinct commit when passing/passing builds

fermadeiral:
This PR intends to fix the issue #490.
A branch generated with the changes contained in this PR can be seen here: https://github.com/fermadeiral/bears-collection-pp/tree/INRIA-spoon-191511078-20170113-031556
The process is right (of committing changed tests in a distinct commit when passing/passing builds), but this branch, for instance, is not completely correct because of the issue #581. However, the two issues are different, thus I propose this PR for adding the feature for issue #490, and in another moment (PR) I will fix the bug reported in issue #581.

PS: One of the last commits is a bug fix (I introduced such bug when I extracted the method `copyDirectory`). Sorry for mixing it in this PR, but I needed to fix that to see if the things were working for the feature added in this PR.

PR #587, opened 2018-05-28
Fix a bug related to copyDirectory method

fermadeiral:
This PR intends to fix a bug that I introduced when I extracted the method `copyDirectory` in a previous refactoring PR. This bug was also fixed in the PR #584, but that one will take some time to be ready, and I need this fix to test properly other stuff I'm doing now.

Well, if the doc in the method `copyDirectory` is good, I don't need to describe what was the bug. Let's see if I need to update the doc...

PS: the build is passing in https://github.com/fermadeiral/repairnator/pull/7

### DIFF
--- a/repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/git/GitHelper.java
+++ b/repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/git/GitHelper.java
@@ -338,40 +338,50 @@ public class GitHelper {
     }
 
     /**
-     * Copy the files from a directory to another.
+     * Copy the files from a directory into another.
      *
      * @param sourceDir is the directory containing the files to be copied from.
      * @param targetDir is the directory where the files from sourceDir are going to be copied to.
-     * @param excludedFileNames is a list of file names not to be included in targetDir.
-     * @param isToRemove when set as "true", all files will be copied from sourceDir to targetDir, and after that files with the names in excludedFileNames will be removed from targetDir.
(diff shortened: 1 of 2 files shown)


repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/git/GitHelper.java
```diff
@@ -337,19 +337,39 @@ public void gitAdd(List<String> files, Git git) {
...
     }
 
-    public void copyDirectory(File sourceDir, File targetDir, String[] excludedFileNames, AbstractStep step) {
+    /**
+     * Copy the files from a directory to another.
+     *
+     * @param sourceDir is the directory containing the files to be copied from.
+     * @param targetDir is the directory where the files from sourceDir are going to be copied to.
+     * @param excludedFileNames is a list of file names not to be included in targetDir.
+     * @param isToRemove when set as "true", all files will be copied from sourceDir to targetDir, and after that files with the names in excludedFileNames will be removed from targetDir.
```
surli: Reading the beginning of the doc it looks like it's about the process of copying/deleting file, but at the end it looks more about the way the previous argument is used.

If I follow you, `isToRemove` is only used to know if the `excludedFileName` should correspond exactly to a fileName or if it should be used as a pattern / a substring of a filename. Right?

fermadeiral: I understand your point, and it means to me I should change stuff. I will do changes in both code and doc and push it again.

Issue #168, opened 2017-03-30
bug: in case a final / remain in push URL, the push will fail


surli:

fermadeiral:
Simon, I don't understand that issue. Could you elaborate on it for me, please?

surli:
The remote repo should be exactly on the form:

`https://github.com/user/repo`

A remote repo like: `https://github.com/user/repo/` or `https://github.com/user/repo.git` can led to a bug when pushing.

PR #589, opened 2018-05-28
Add check for pushurl arg

fermadeiral:
This PR intends to fix issue #168.
I thought on two ways to fix the mentioned issue:
1) To try to repair the push url in the push step (last step of the pipeline)
2) To check the correctness of the push url when other variables are also checked (before the pipeline starts)

I took the second way, since the first way might not work depending on how wrong the push url is, and the user of repairnator would know there was a problem at pushing only in the end of the execution.
WDYT?

PS: the build is passing in https://github.com/fermadeiral/repairnator/pull/8

### DIFF
--- a/repairnator/repairnator-core/src/test/java/fr/inria/spirals/repairnator/TestUtils.java
+++ b/repairnator/repairnator-core/src/test/java/fr/inria/spirals/repairnator/TestUtils.java
@@ -15,6 +15,13 @@ public class TestUtils {
     }
 
     @Test
+    public void testMatchesGithubRepoUrlWithRightRepoUrlWithNumber() {
+        String repoUrl = "https://github.com/Sp1r4ls-T34m/r3p41rn4t0r";
+
+        assertThat(Utils.matchesGithubRepoUrl(repoUrl), is(true));
+    }
+
+    @Test
     public void testMatchesGithubRepoUrlWithWrongRepoUrl() {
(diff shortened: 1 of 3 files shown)


repairnator/repairnator-core/src/test/java/fr/inria/spirals/repairnator/TestUtils.java
```diff
@@ -0,0 +1,31 @@
...
+
+    @Test
+    public void testMatchesGithubRepoUrlWithWrongRepoUrl() {
+        String repoUrl = "https://github.com/Spirals-Team/repairnator/";
+
+        assertThat(Utils.matchesGithubRepoUrl(repoUrl), is(false));
+    }
+
+    @Test
+    public void testMatchesGithubRepoUrlWithWrongRepoUrl2() {
```
surli: Could you add a test with project containing numbers?

fermadeiral: Sure. It's done, and it works for numbers.

However, I just realized that repo name can contain dot, and also I checked the rule for user name ("Username may only contain alphanumeric characters or single hyphens, and cannot begin or end with a hyphen and is already taken" from GitHub). I will do some changes for these things. I tell you when it's done.

fermadeiral:
@surli this PR is ready again.

surli:
Great thanks!

Issue #588, opened 2018-05-28
security: github token might be disclosed if the command line is dumped


surli:
Some project might even dumped the command line of the current running process which contain the github token: in that case, the token might appear in a log and be pushed in a repository.

In order to avoid that, we should analyze the produced logs before they are pushed and remove all occurence of the pair argument + token.

We should not only check for the appearance of the token as it might be a way to disclose a token.

surli:
Fixed by #590

PR #590, opened 2018-05-28
Remove ghOauth from created files to push

fermadeiral:
This PR intends to fix issue #588.

I run repairnator to REPAIR the build 384520172, which was the build that exposed the bug. The changes work, mainly because I didn't receive an email from GitHub after the branch was generated: https://github.com/fermadeiral/bears-collection-fp/tree/marlon123-spring-api-rest-curso-384520172-20180527-202310

PS: the build is passing in https://github.com/fermadeiral/repairnator/pull/9

### DIFF
--- a/repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/git/GitHelper.java
+++ b/repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/git/GitHelper.java
@@ -41,6 +41,8 @@ import java.io.File;
 import java.io.FileFilter;
 import java.io.FileWriter;
 import java.io.IOException;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/git/GitHelper.java
```diff
@@ -404,4 +406,27 @@ public void removeNotificationFromTravisYML(File directory, AbstractStep step) {
...
+        for (String fileName : fileNames) {
+            File file = new File(directory, fileName);
+
+            if (!file.exists()) {
+                getLogger().warn("The file "+file.toPath()+" does not exist.");
+            } else {
+                Charset charset = StandardCharsets.UTF_8;
+                try {
+                    String content = new String(Files.readAllBytes(file.toPath()), charset);
+                    String updatedContent = content.replaceAll(ghOauthPattern, "");
```
surli: Maybe it's better to add a tag "[REMOVED]" instead of nothing.

repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/git/GitHelper.java
```diff
@@ -404,4 +406,27 @@ public void removeNotificationFromTravisYML(File directory, AbstractStep step) {
...
+        String ghOauthPattern = "--ghOauth\\s+[\\w]+";
+        for (String fileName : fileNames) {
+            File file = new File(directory, fileName);
+
+            if (!file.exists()) {
+                getLogger().warn("The file "+file.toPath()+" does not exist.");
+            } else {
+                Charset charset = StandardCharsets.UTF_8;
+                try {
+                    String content = new String(Files.readAllBytes(file.toPath()), charset);
```
surli: I really don't know: do you know if it's the most efficent way of doing it? We might be manipulating logs of big size (more than 50 Mb in some cases). I just want to be sure it's the good way to do it.

repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/git/GitHelper.java
fermadeiral: Done.

repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/git/GitHelper.java
fermadeiral: I also don't know. This was the simpler way I found that does the job, but I don't know if this is also the most efficient way. I will do some research, but I promise nothing because I don't know if I will find what we want to know.

surli: Based on this answer: https://stackoverflow.com/questions/326390/how-do-i-create-a-java-string-from-the-contents-of-a-file?rq=1 it looks like the good way to do it. I'll merge the PR.

Issue #581, opened 2018-05-22
git.checkout().setStartPoint(commitCheckout).addPaths(paths).call() seems not to work as expected.


fermadeiral:
When trying to solve #490, I got a case when `git.checkout().setStartPoint(commitCheckout).addPaths(paths).call()` seems not to work as we expected.

Example: https://github.com/INRIA/spoon/commit/551f0fb28c49f4b1eeab93f7748c04ef82cfde39 (this commit is a fixer commit).

There are 3 test classes involved:
1) src/test/java/spoon/test/imports/ImportTest.java
2) src/test/java/spoon/test/imports/testclasses/TransportIndicesShardStoresAction.java (NEW)
3) src/test/java/spoon/test/imports/testclasses2/AbstractMapBasedMultimap.java

Consider I have checked out that commit.
Then I have its previous commit called `commitCheckout`, and a variable `paths` containing `src/test/java`.
When I call `git.checkout().setStartPoint(commitCheckout).addPaths(paths).call()`, files 1 and 3 are updated for their previous state, but file 2, the new one, remains checked out.

I see no other solution than discovering the new files and deleting them by force. I also still don't know if the opposite happens (i.e., if a deleted file is not checked out in the commit it existed).

surli:
okay... It might be another behaviour of JGit which is not standard compared to the git one -_-

fermadeiral:
I talked with Thomas about that issue since his knowledge on git is advanced and in fact he showed me that `git checkout <paths>` is not the right command to do the job I want, i.e. the problem is not with JGit. He gave me a sequence of commands that does the job properly:

```
git reset <commit> -- <paths>
git clean -fd
git checkout -- .
```
Thanks @tdurieux 

@surli, a PR is coming.

surli:
Fixed in #592

PR #592, opened 2018-05-30
Fix checkout files in passing/passing case for Bears

fermadeiral:
This PR fixes issue #581.

Since you told me that you intend to remove JGit from repairnator, I do not use JGit for the new git commands done in order to reduce work in the future.

PS: the build is passing in https://github.com/fermadeiral/repairnator/pull/10


surli:
Test are passing but did you try this one and check the content of pushed info?

fermadeiral:
> Test are passing but did you try this one and check the content of pushed info?

Oh, I forgot to mention. I tested these changes from inside the changes in PR #584, running with the pair of builds that the example given in issue #581 was made up, and the generated branch is what it should be...I also checked the console messages in every step child of `CheckoutRepository` and followed the output from those new git commands and that worked as expected.

surli:
Great then

PR #593, opened 2018-05-30
Add bears.json

fermadeiral:
This PR adds the creation of the `bears.json` for all generated branches from repairnator. Independently of repairnator being run to repair or to collect bugs, both `repairnator.json` and `bears.json` are created.
Here I show you some examples of `bears.json` being generated with the changes proposed in this PR:

- REPAIR, not fork, not PR (without info on fixer build, fork is false, PR is false, one commit):
https://github.com/fermadeiral/bears-collection-fp/blob/surli-test-repairnator-352395977-20180312-123748/bears.json

- REPAIR, fork, not PR (without info on fixer build, fork is true, PR is false, one commit):
https://github.com/fermadeiral/bears-collection-fp/blob/danglotb-dspot-378658748-20180514-074545/bears.json

- BEARS, not fork, PR (with info on fixer build, fork is false, PR is true, 6 commits):
https://github.com/fermadeiral/bears-collection-fp/blob/INRIA-spoon-369997088-20180423-055639/bears.json

Opened issues regarding bears.json schema:
- Some properties do not perfectly match the schema because I renamed some of them: I will make them match after we merge the PR on the schema;
- In this current PR, the info on step durations are missing.

Even the `bears.json` not being 100%, I would say it is already 95% of the job done, thus I see no problem at this PR entering to review to be merged (this PR involves too much files, and I would like to merge it as soon as possible to avoid conflicts with my next PRs).

Issues to be fixed (I will created issues in repairnator repo and propose PRs to address them after this PR being merged):
- Patch stats are buggy (they already were before the changes done here) and I already know why and how to fix it;
- `version` property is hard-coded and I will change that.

PS: the build is passing in https://github.com/fermadeiral/repairnator/pull/6


surli:
I propose to merge this one only when #499 is merged.
And could you add a new unit test checking the validation of the created bears.json file against the schema. You can use this project in order to dot that: https://github.com/java-json-tools/json-schema-validator ?

surli:
Why did you close this one?

fermadeiral:
I will open it again when it's 100% ready to go.

surli:
ok then

fermadeiral:
> And could you add a new unit test checking the validation of the created bears.json file against the schema. You can use this project in order to dot that: https://github.com/java-json-tools/json-schema-validator ?

Quick question: the json file can be in different ways: 1) with PR info; 2) with fork info; 3) with buggy build only; 4) with buggy and fixer builds; 5) the 1 with the 3, 6)... (this is why I had shown you more than one generated json file in the description of this PR). So, for the test, do I put a test for each case (which will make the test suite slower to be run) or do I choose some cases? If you pick the second option, which cases do you suggest me to put in the tests?

surli:
Try 3 and 4. In the worst case we can always ignore some tests.

fermadeiral:
Ok. I will start with 3 and 4.

PR #594, opened 2018-05-31
WIP Fix bug in MavenModelResolver

surli:
Some times it appears that the `SOURCEDIRNOTCOMPUTED` status happens because we do not manage to resolve properly the maven model. This PR tries to fix that issue.


fermadeiral:
Is this bug still an existing bug?

surli:
Yes the bug still exists.

fermadeiral:
Ok ;/

Do you intend to open an issue for that since it seems there is no fix for that right now?

surli:
Done

PR #608, opened 2018-06-12
Refactor minor things

fermadeiral:
This PR proposes minor refactorings, mainly related to constants and small util methods.
The build is passing in https://github.com/fermadeiral/repairnator/pull/15


repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/pipeline/Launcher.java
```diff
@@ -355,10 +355,10 @@ private void mainProcess() {
 
         ProjectInspector inspector;
 
-        if (config.getLauncherMode() == LauncherMode.BEARS) {
-            inspector = new ProjectInspector4Bears(buildToBeInspected, this.config.getWorkspacePath(), serializers, this.notifiers);
-        } else {
+        if (config.getLauncherMode() == LauncherMode.REPAIR) {
```
surli: Why did you change that if? It seems to me more interesting to check if it's bears or not: the default case should be repair...

fermadeiral: Because in line 344 there is an if for `LauncherMode.REPAIR`. I just wanted to make the two ifs to match to be easier to understand the code.

fermadeiral: I changed the if back, and I put the other if comparing with `LauncherMode.BEARS` then.

PR #617, opened 2018-06-13
Small changes to speed up just a little bit the test suite execution

fermadeiral:
The build is passing in https://github.com/fermadeiral/repairnator/pull/19

### DIFF
--- a/repairnator/repairnator-pipeline/src/test/java/fr/inria/spirals/repairnator/process/step/paths/TestComputeClasspath.java
+++ b/repairnator/repairnator-pipeline/src/test/java/fr/inria/spirals/repairnator/process/step/paths/TestComputeClasspath.java
@@ -4,8 +4,8 @@ import ch.qos.logback.classic.Level;
 import fr.inria.jtravis.entities.Build;
 import fr.inria.spirals.repairnator.BuildToBeInspected;
 import fr.inria.spirals.repairnator.process.inspectors.StepStatus;
+import fr.inria.spirals.repairnator.process.step.BuildProject;
 import fr.inria.spirals.repairnator.process.step.CloneRepository;
-import fr.inria.spirals.repairnator.process.step.TestProject;
 import fr.inria.spirals.repairnator.process.step.checkoutrepository.CheckoutType;
 import fr.inria.spirals.repairnator.states.ScannedBuildStatus;
 import fr.inria.spirals.repairnator.Utils;
@@ -85,7 +85,7 @@ public class TestComputeClasspath {
         ComputeClasspath computeClasspath = new ComputeClasspath(inspector, true);
 
         cloneStep.setNextStep(new CheckoutBuggyBuild(inspector, true))
-                .setNextStep(new TestProject(inspector))
+                .setNextStep(new BuildProject(inspector))
                 .setNextStep(computeClasspath);
         cloneStep.execute();
 
--- a/repairnator/repairnator-pipeline/src/test/java/fr/inria/spirals/repairnator/process/step/paths/TestComputeClasspath.java
+++ b/repairnator/repairnator-pipeline/src/test/java/fr/inria/spirals/repairnator/process/step/paths/TestComputeClasspath.java
@@ -4,7 +4,6 @@ import ch.qos.logback.classic.Level;
 import fr.inria.jtravis.entities.Build;
 import fr.inria.spirals.repairnator.BuildToBeInspected;
 import fr.inria.spirals.repairnator.process.inspectors.StepStatus;
-import fr.inria.spirals.repairnator.process.step.BuildProject;
 import fr.inria.spirals.repairnator.process.step.CloneRepository;
 import fr.inria.spirals.repairnator.process.step.checkoutrepository.CheckoutType;
 import fr.inria.spirals.repairnator.states.ScannedBuildStatus;
@@ -85,14 +84,13 @@ public class TestComputeClasspath {
         ComputeClasspath computeClasspath = new ComputeClasspath(inspector, true);
 
         cloneStep.setNextStep(new CheckoutBuggyBuild(inspector, true))
-                .setNextStep(new BuildProject(inspector))
                 .setNextStep(computeClasspath);
         cloneStep.execute();
 
         assertThat(computeClasspath.isShouldStop(), is(false));
         List<StepStatus> stepStatusList = jobStatus.getStepStatuses();
-        assertThat(stepStatusList.size(), is(4));
-        StepStatus classpathStatus = stepStatusList.get(3);
+        assertThat(stepStatusList.size(), is(3));
+        StepStatus classpathStatus = stepStatusList.get(2);
         assertThat(classpathStatus.getStep(), is(computeClasspath));
 
         for (StepStatus stepStatus : stepStatusList) {
@@ -111,6 +109,11 @@ public class TestComputeClasspath {
         expectedClasspath.add(junit);
         expectedClasspath.add(hamcrest);
 
-        assertThat(jobStatus.getRepairClassPath(), is(expectedClasspath));
+        List<URL> actualClasspath = new ArrayList<>();
+        actualClasspath.add(classDir);
+        actualClasspath.add(testDir);
+        actualClasspath.addAll(jobStatus.getRepairClassPath());
+
+        assertThat(actualClasspath, is(expectedClasspath));
     }
 }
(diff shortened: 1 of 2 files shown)


repairnator/repairnator-pipeline/src/test/java/fr/inria/spirals/repairnator/process/step/paths/TestComputeClasspath.java
```diff
@@ -85,7 +85,7 @@ public void testComputeClasspath() throws IOException {
         ComputeClasspath computeClasspath = new ComputeClasspath(inspector, true);
 
         cloneStep.setNextStep(new CheckoutBuggyBuild(inspector, true))
-                .setNextStep(new TestProject(inspector))
+                .setNextStep(new BuildProject(inspector))
```
surli: I'm not sure: why did you keep this step here? We should not need it at all.

fermadeiral: This was what I thought too, but it didn't work. So I changed for one that normally takes less time to execute at least.

surli: Any idea why it did not work?

fermadeiral: The only reason I can think right now is because the folder "target" is not available without maven step, which is expected in `ComputeClasspath`.

surli: You're right because I put target/classes and target/test-classes. So we should change the test to create the folder before calling the step and it should work properly.

fermadeiral:
Ready for review again.

repairnator/repairnator-pipeline/src/test/java/fr/inria/spirals/repairnator/process/step/paths/TestComputeClasspath.java
```diff
@@ -111,6 +109,11 @@ public void testComputeClasspath() throws IOException {
         expectedClasspath.add(junit);
         expectedClasspath.add(hamcrest);
 
-        assertThat(jobStatus.getRepairClassPath(), is(expectedClasspath));
+        List<URL> actualClasspath = new ArrayList<>();
+        actualClasspath.add(classDir);
+        actualClasspath.add(testDir);
```
surli: This is not what I had in mine. I was thinking about creating manually the directories:

```
File classDir = new File(repoDir.getAbsolutePath(), "target/classes");
classDir.mkdir();

File testDir = new File(repoDir.getAbsolutePath(), "target/classes");
testDir.mkdir();
```

And then to check the result is good.

fermadeiral: Could you tell me why this way is better?

Supposing I agree with you, so what you are saying to me is to "break" the execution of the steps in the middle to create the directories?
It's like this now:
```
cloneStep.setNextStep(new CheckoutBuggyBuild(inspector, true))
                 .setNextStep(computeClasspath);
         cloneStep.execute();
```
So what you want is this:
```
cloneStep.setNextStep(new CheckoutBuggyBuild(inspector, true));
         cloneStep.execute();

// create directories

computeClasspath.execute();
```
?

surli: It don't think it's needed to break the execution. The way the test is built, the project is cloned exactly at path repoDir, so it should be ok to create before the cloning step the directories for repoDir/target/classes and test-classes.

fermadeiral: You didn't answer my first question, but I gave a shot for this even without understanding why I should do in your way, and it didn't work:

`11:56:03.323 [main] ERROR f.i.s.r.process.step.CloneRepository - Destination path "repo" already exists and is not an empty directory`

surli: > Could you tell me why this way is better?

I check the contract that `target/classes` and `target/test-classes` are added in the classpath. Here you're changing the output of the program to make it acceptable regarding the expected: it would be even better to change directly the expected to remove `classes` and `test-classes`.

Yeah I kind of expected this error to come up. Remove this check in CloneRepository: only warn the user. It should never happen for us as we are using it in a docker container.

fermadeiral: > I check the contract that target/classes and target/test-classes are added in the classpath. Here you're changing the output of the program to make it acceptable regarding the expected: it would be even better to change directly the expected to remove classes and test-classes.

Thanks for clarifying.

fermadeiral:
@surli, you have said "Remove this check in CloneRepository: only warn the user.". What about to warn the user plus to delete the existing directory, so the clone would be from scratch for sure, always?

fermadeiral:
Ok then. I'm in this issue now because I believe that any opportunity we already know that might speed up the test suite execution is high priority, since we always have to wait for Travis.

fermadeiral:
> However a middle way might be to rename the existing directory with a timestamp and to warn the user about this renaming.

This is not working =/

The issue in the test `testComputeClasspathWithSingleModuleProject()` is the following:
1) first we have a tmpDir, i.e. `test_computecp`
2) then we create the repoDir, i.e. `test_computecp/repo`
3) then we create the targetDir, i.e. `test_computecp/repo/target`
4) then we create the classes and test-classes dirs, i.e. `test_computecp/repo/target/classes` and `test_computecp/repo/target/test-classes`
5) when the step `CloneRepository` is called, I rename the existing dir (i.e. `test_computecp/repo`) like this:

```
this.getLogger().debug("Cloning repository " + repoSlug + " in the following directory: " + repoLocalPath);

+ File repoLocal = new File(repoLocalPath);
+ if (repoLocal.exists() && repoLocal.isDirectory()) {
+    String fileNameDate = Utils.formatFilenameDate(new Date());
+    File newNameForOldLocalRepo = new File(repoLocal.getAbsolutePath() + "-" + fileNameDate);
+    this.getLogger().warn("Destination path " + repoLocalPath + " already exists. It will be renamed to "
+                        + newNameForOldLocalRepo.getName());
+    repoLocal.renameTo(newNameForOldLocalRepo);
+ }

Git.cloneRepository().setCloneSubmodules(true).setURI(repoRemotePath).setDirectory(new File(repoLocalPath)).call();
```

The previous error "[main] ERROR f.i.s.r.process.step.CloneRepository - Destination path "repo" already exists and is not an empty directory" does not occur anymore when we call jgit command, but the assertion fails since the `target/classes` and `target/test-classes` dirs are in the renamed folder and the cloned repo is in the new folder. Any other suggestion?

fermadeiral:
I'm a little bit confused if I should open the new PR on this now, even it's not working, for you to see the changes I'm performing, so tell me if you prefer that.

surli:
> This is not working =/

Okay here we should distinguish two things: 
  1. the expected behaviour of the program; 
  2. the unit test we are working one.

so in 1 we want to avoid overwritting or deleting existing data.
In 2. we want to ensure that a path contain the right data: we don't care actually about the name of the first repo, so we can create a randomly named directory at first. Or a directory based on the timestamp: then when launching multiple times the tests it shouldn't be a problem, should it?

fermadeiral:
> Okay here we should distinguish two things:

    the expected behaviour of the program;
    the unit test we are working one.

Agreed.

> In 2. we want to ensure that a path contain the right data: we don't care actually about the name of the first repo, so we can create a randomly named directory at first. Or a directory based on the timestamp: then when launching multiple times the tests it shouldn't be a problem, should it?

Ok, I think we are not talking about the same thing, so let me try to explain the issue. I ask you now to imagine the test being executed only one time, because the issue is on it.

1) The step `ComputeClasspath` computes the jars and also the paths `target/classes` and `target/test-classes`: these info compose the classpath, so this is what the test on such step should check.

2) However, without a maven step, only the jars are computed, since `target/classes` and `target/test-classes` don't exist.

3) The test we are discussing executes the step `TestProject`, so the `target/classes` and `target/test-classes` dirs are created, and the test passes just fine.

4) However, we don't want to execute the step `TestProject`, and without it, we are in the problem described in 2.

5) The solution you proposed to me in the past was to create the dirs `target/classes` and `target/test-classes` from the test, before cloning the repo from the build being tested.

6) The REAL problem here then is not about launching multiple times the test: the problem is that a dir is created with `target/classes` and `target/test-classes` by the test, and then when `CloneRepository` is executed, the project is cloned in a different folder (since the other one already exists with the `target` dir), so we "lose" the `target` folders, which makes the test to fail anyways because only the jars are returned by `ComputeClasspath`.

fermadeiral:
I proposed the mentioned solution in PR #666

Issue #610, opened 2018-06-12
Project metrics seem buggy.


fermadeiral:
We have three project metrics:
- numberSourceFiles (calculate in `ComputeSourceDir`)
- numberTestFiles (calculate in `ComputeTestDir`)
- numberLibraries (calculate in `ComputeClasspath`)

The correctness of these metrics is not being checked, and they seem buggy based on values generated by repairnator in `bears.json` file.

fermadeiral:
Contract for `numberLibraries` (agreement in #624): for Repairnator, it's on the failing module (it's already correct); for Bears, it's on all modules.

surli:
Put it as documentation in the code, please.

fermadeiral:
The `numberLibraries` calculation for Bears will remain the same as for Repairnator: it's on the failing module.

surli:
fixed in #656

PR #618, opened 2018-06-14
Fix numberSourceFiles metric

fermadeiral:
This PR contributes with issue #610 by fixing the `numberSourceFiles` metric (related to `ComputeSourceDir` step).

This PR fixes two bugs:
- The source dirs had been got twice in `ComputeSourceDir` step, and by consequence the `numberSourceFiles` metric had been set with the double of the actual number of source files. This was fixed in `ComputeSourceDir`.
- Specific for `ProjectInspector` (REPAIR), the resulted `numberSourceFiles` metric on json files was always 0, because the json files had been created before the final steps being executed, and `ComputeSourceDir` was one of them. To fix that, I created the step `WritePropertyFile` to be responsible for writing the json files, which is a final step, just before `CommitProcessEnd`. I also moved `properties` from `AbstractStep` to `JobStatus` to make the things work properly (because this info is not passed to final steps since the first final step doesn't have a previous step). This actually made the code in `AbstractStep` cleaner and less complex IMHO.

The build is passing in https://github.com/fermadeiral/repairnator/pull/20


repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/step/WritePropertyFile.java
```diff
@@ -0,0 +1,77 @@
...
+                file.createNewFile();
+            }
+            OutputStreamWriter outputStream = new OutputStreamWriter(new FileOutputStream(file));
+            outputStream.write(jsonString);
+            outputStream.flush();
+            outputStream.close();
+            this.getInspector().getJobStatus().addFileToPush(file.getName());
+            return true;
+        } catch (IOException e) {
+            this.getLogger().error("An exception occurred when writing the following property file: " + file.getPath(), e);
```
surli: You could use addStepError here.

repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/step/gatherinfo/GatherTestInformation.java
```diff
@@ -122,12 +125,12 @@ public FileVisitResult preVisitDirectory(Path file, BasicFileAttributes attrs) t
                         this.failingModulePath = failingModule.getCanonicalPath();
 
                         if (!this.skipSettingStatusInformation) {
-                            this.getInspector().getJobStatus().setFailingModulePath(this.failingModulePath);
-                            this.writeProperty("failingModule", this.failingModulePath);
-                            this.getInspector().getJobStatus().getMetrics4Bears().getTests().setFailingModule(this.failingModulePath);
+                            jobStatus.setFailingModulePath(this.failingModulePath);
```
surli: Actually we could even move the two lines below in the `setFailingModulePath` method.

repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/step/gatherinfo/GatherTestInformation.java
```diff
@@ -194,23 +197,23 @@ public FileVisitResult preVisitDirectory(Path file, BasicFileAttributes attrs) t
...
-            this.writeProperty("error-types", this.failureNames);
-            this.writeProperty("failing-test-cases", this.failureLocations);
-            this.writeProperty("totalNumberFailingTests", this.nbFailingTests);
-            this.writeProperty("totalNumberErroringTests", this.nbErroringTests);
-            this.writeProperty("totalNumberSkippingTests", this.nbSkippingTests);
-            this.writeProperty("totalNumberRunningTests", this.nbTotalTests);
-            this.getInspector().getJobStatus().setFailureLocations(this.failureLocations);
-
-            Metrics metrics = this.getInspector().getJobStatus().getMetrics();
+            jobStatus.writeProperty("error-types", this.failureNames);
```
surli: Same comment as below: we could move the write properties in the setters of JobStatus.

fermadeiral: For the comment below, I agree, but here I don't really understand, since `JobStatus` does not have setters for these stuff. Is supposed for me to create them?

surli: No actually I was saying that because they are in metrics. But for now leave it as it is, add just a todo comment to fix that later.

fermadeiral: Just for the record, actually I think that this `writeProperty` method will disappear in the future when `repairnator.json` is refactored to close to what `bears.json` is, because for `bears.json` we don't use this method (everything is in `Metrics4Bears`).

fermadeiral: Do you want I remove the duplication instead of putting the TODO comment? If yes, what do I remove? `jobStatus.writeProperty` or `metrics.set`? See that only for `failureLocations` we need the `jobStatus.writeProperty`, because apparently this property doesn't exist in `Metrics`.

surli: yes keep the failure location remove the rest

fermadeiral: Ok.

fermadeiral:
I'm waiting the build to finish in my fork.

fermadeiral:
The build is passing, this PR is ready again.

PR #619, opened 2018-06-14
Clean pipeline tests

fermadeiral:
The build is passing in https://github.com/fermadeiral/repairnator/pull/21


surli:
This is a long diff here, it's only about test refactoring? You did not change any assert right?

fermadeiral:
> This is a long diff here, it's only about test refactoring?

Mainly. The only change that changes the behavior is that before the tmp dirs were not deleted, but now they are.

> You did not change any assert right?

In theory I did change because the new method `checkBuildAndReturn` in the end of almost all test classes performs a more "complete" assertion than it had been done in some classes. For instance, in `TestCloneRepositoryStep` there was no `assertThat(build.isPullRequest(), Is.is(isPR));`, but now it has. But regarding changes in assertions, this was all that I did.

PR #622, opened 2018-06-14
Fix NPE in scanner launcher

fermadeiral:
This PR fixes a NPE in scanner launcher (I probably introduced that in some refactoring).


repairnator/repairnator-scanner/src/main/java/fr/inria/spirals/repairnator/scanner/Launcher.java
```diff
@@ -61,14 +61,14 @@ public Launcher(String[] args) throws JSAPException {
         JSAP jsap = this.defineArgs();
         JSAPResult arguments = jsap.parse(args);
         LauncherUtils.checkArguments(jsap, arguments, LauncherType.SCANNER);
+        this.initConfig(arguments);
```
surli: Could you add a comment here to specify that "this.config" will be available only after this call. It might avoid future NPE here.

fermadeiral: Done.

Issue #554, opened 2018-05-15
skipDelete for dockerpool should be settable in repairnator.cfg


surli:
skipDelete is only a Java code argument, we should be able to set it from outside

monperrus:
Benjamin is working on this and has the branch to get the WIP PR open

surli:
fixed in #642

PR #642, opened 2018-06-28
review Skip delete option

btellstrom:
I have started to work on #554 at the moment it does however not appear to be enough. When I attempt to run repair_buggy_build.sh I receive the following: 
```
Launch docker container with the following command: docker run -d --env REPAIR_MODE=repair --env BUILD_ID=352395977 --env GITHUB_OAUTH=* --env LOG_FILENAME=repairnator-pipeline_2018-06-28_1701_352395977 --env RUN_ID=ec3b867c-f9d3-47ee-a22d-b49c55736378 --env OUTPUT=/var/log --env REPAIR_TOOLS=Nopol,AstorJMut,NPEFix --env SKIP_DELETE=1
```
which is as it should, but also this:
```
with the following config: RepairnatorConfig{runId='ec3b867c-f9d3-47ee-a22d-b49c55736378', launcherMode=REPAIR, serializeJson=true, inputPath='null', outputPath='/var/log', mongodbHost='null', mongodbName='repairnator', smtpServer='null', notifyTo=[], notifyEndProcess=false, push=false, pushRemoteRepo='null', fork=false, lookFromDate=null, lookToDate=null, buildId=352395977, z3solverPath='./z3_for_linux', workspacePath='./workspace', githubToken='*', dockerImageName='null', skipDelete=false, createOutputDir=false, logDirectory='null', nbThreads=0, globalTimeout=0, whiteList=null, blackList=null, jobSleepTime=0, buildSleepTime=0, maxInspectedBuilds=0, duration=null, humanPatch=false, repository='null', clean=true, bearsMode=BOTH, bearsDelimiter = false, repairTools=Nopol,NPEFix,AstorJMut}

```
which clearly says skipDelete=false, and the containers have been deleted after the run. Will continue to work on it, if anyone has any suggestions in the meantime I'd appreciate them.


surli:
Actually that's normal: the argument should be used when calling dockerpool library: this is used when you have severals docker container. When you use `repair_buggy_build` we only create one container, so we don't use dockerpool.

You can see that we call: 
```
DOCKER_COMMAND="docker run -d $DOCKER_ARGS -v $LOG_DIR:/var/log $DOCKER_TAG"
```

the `-d` means that we always delete the docker container after `repair_buggy_build`. So here you should check if you need to put this argument or not.

repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/pipeline/Launcher.java
```diff
@@ -158,6 +158,11 @@ private JSAP defineArgs() throws JSAPException {
         opt2.setRequired(true);
         opt2.setDefault(repairTools);
         jsap.registerParameter(opt2);
+        
+        Switch sw1 = new Switch("skipDelete");
+        sw1.setLongFlag("skipDelete");
+        sw1.setDefault("false");
+        jsap.registerParameter(sw1);
```
fermadeiral: If this is really necessary, you should do this way:
```
// --skipDelete
jsap.registerParameter(LauncherUtils.defineArgSkipDelete());
```
Because this arg definition already exists.

Also, in the method `initConfig` of such class, you should do this:

`this.config.setSkipDelete(LauncherUtils.getArgSkipDelete(arguments));`

To set the arg value in `RepairnatorConfig`.

btellstrom: Fixed

fermadeiral: Thank you, Benjamin. But I'm not sure it's necessary to change this class. Simon will probably check it before merging anyway.

repairnator/repairnator-checkbranches/src/main/java/fr/inria/spirals/repairnator/checkbranches/Launcher.java
```diff
@@ -77,6 +77,8 @@ private JSAP defineArgs() throws JSAPException {
         jsap.registerParameter(LauncherUtils.defineArgNbThreads());
         // -g or --globalTimeout
         jsap.registerParameter(LauncherUtils.defineArgGlobalTimeout());
+        // --skipDelete
+        jsap.registerParameter(LauncherUtils.defineArgSkipDelete());
```
fermadeiral: In this class, this already exists.

repairnator/repairnator-dockerpool/src/main/java/fr/inria/spirals/repairnator/dockerpool/Launcher.java
```diff
@@ -92,6 +92,8 @@ private JSAP defineArgs() throws JSAPException {
         jsap.registerParameter(LauncherUtils.defineArgPushUrl());
         // --ghOauth
         jsap.registerParameter(LauncherUtils.defineArgGithubOAuth());
+        // --skipDelete
+        jsap.registerParameter(LauncherUtils.defineArgSkipDelete());
```
fermadeiral: In this class, this already exists too.

btellstrom: Fixed, noticed that as well.

repairnator/scripts/check_branches.sh
```diff
@@ -58,6 +58,9 @@ fi
 if [ "$HUMAN_PATCH" -eq 1 ]; then
     args="$args --humanPatch"
 fi
+if [ "$SKIP_DELETE" -eq 1 ]; then
+    elementaryArgs="$elementaryArgs --skipDelete"
```
fermadeiral: `args` instead of `elementaryArgs`

repairnator/scripts/launch_rtscanner.sh
```diff
@@ -47,5 +47,9 @@ if [ "$CREATE_OUTPUT_DIR" -eq 1 ]; then
     args="$args --createOutputDir"
 fi
 
+if [ "$SKIP_DELETE" -eq 1 ]; then
+    elementaryArgs="$elementaryArgs --skipDelete"
```
fermadeiral: `args` instead of `elementaryArgs`

fermadeiral:
I'm not sure what you boys had discussed, but I don't think that the new option should be add in `scanner/Launcher` and `launch_scanner.sh`. This is the only module we have that does not use docker at all.

Moreover, I'm not sure if it's necessary to create in `pipeline/Launcher` such option too. Please, confirm this for me.

repairnator/repairnator-dockerpool/src/main/java/fr/inria/spirals/repairnator/dockerpool/Launcher.java
fermadeiral: :-)

btellstrom:
That is most likely just me not being used to docker and uncertain where it is used and where it is not used. I made the assumption we used docker for everything java-related. 

Also, thank you for the feedback, although I am slightly embarrassed by the script errors...

repairnator/scripts/check_branches.sh
btellstrom: Fixed.

repairnator/scripts/launch_rtscanner.sh
btellstrom: Fixed.

fermadeiral:
> That is most likely just me not being used to docker and uncertain where it is used and where it is not used. I made the assumption we used docker for everything java-related.

Which means that we probably have an issue with the documentation of the modules. I will work on that in the future.

> Also, thank you for the feedback, although I am slightly embarrassed by the script errors...

There is no need at all to be embarrassed.

btellstrom:
@surli does it? When looking at 
```
docker run --help
```
it says 
```
  -d, --detach                         Run container in background and print container ID

```
Also, apparently putting it in a conditional as I have done breaks the script, need to fix that.

fermadeiral:
> Also, apparently putting it in a conditional as I have done breaks the script, need to fix that.

The conditional seems fine. It can be a lack of space between "-d" and what has been added in the args after.

btellstrom:
You were correct. Also forgot to add $DOCKER_ARGS to the next line. However, that just removed the -d without a conditional. It was not breaking the script, only that it displayed nothing but a compilation error, compared to the regular terminal output.

surli:
Lot of discussion here! 
So yeah, sorry @btellstrom I was too fast when reading the script and indeed `-d` is not used to delete the container.
Now when you say: `the containers have been deleted after the run`. Actually it currently should not happen with this script: did you check by typing `docker ps -a`. The standard command `docker ps` only display the currently running docker containers.

And for now I'd say to keep this behaviour for this script, as its main purpose is to show repairnator for new users. 

And as @fermadeiral mentioned: the `skipDelete` attribute is useless for the launcher of pipeline. It should only be used for `dockerpool`.

btellstrom:
Apparently I still need to learn docker properly, I ran docker ``` docker container ls ``` which supposedly lists all containers. Running ```docker ps -a``` showed that at least for the ```repair_buggy_build.sh``` the containers are not deleted. However, the logs still show ```skipDelete=false```, 

Okay, I'll remove the option from pipeline then.

surli:
> However, the logs still show skipDelete=false,

Ok here that's another story and not an easy one ;)
If you did not change anything in `repairnator.cfg` the current docker image you use for repairnator is `surli/repairnator:latest`. This means that you do not use your own change made in the dockerfile but you use the docker image pushed after the last change on master branch of this repository.
You can check that pretty easily: `docker ps -a` display the last running container, you copy the container ID that interests you and then you type something like: 

`docker logs <container ID> |less` you should see on the top something like: 
> 02:34:49.050 [main] INFO  f.i.s.repairnator.pipeline.Launcher - PIPELINE VERSION: 2.2-SNAPSHOT-2018-05-31 13:04

Then you know immediately when the image has been built.
So now, how do you check your own image? It's a bit of a work I must confess.

First thing to do: `mvn package` on the `repairnator-pipeline` to obtain a `repairnator-pipeline-[...]-with-dependencies.jar` -> you rename it `repairnator-pipeline.jar` and you put it in the directory I just added on the repo: `repairnator/docker-images/pipeline-dockerimage-dev/`.

Then you go to this repo and you type: `docker build .` it will create a new image and give you its id: you copy this ID in `repairnator.cfg` instead of `surli/repairnator:latest` and now you're using the new docker image!

Yeah I know it's really tedious but I did not find any other good dev process for managing docker containers for now :-/

btellstrom:
Okay, I think I am getting slightly confused now. 

I have been using ```surli/repairnator:latest``` in my config file, but still I see that there is a container left behind after starting both ```launch_rtscanner.sh``` and ```repair_buggy_build.sh```. Is skipDelete default behaviour in repairnator? Or at least for pipeline?

Also, you are talking about pipeline which shouldn't have the skipDelete, should I instead do 
>First thing to do: mvn package on the repairnator-pipeline to obtain a repairnator-pipeline-[...]-with-dependencies.jar -> you rename it repairnator-pipeline.jar and you put it in the directory I just added on the repo: repairnator/docker-images/pipeline-dockerimage-dev/.
>
>Then you go to this repo and you type: docker build . it will create a new image and give you its id: you copy this ID in repairnator.cfg instead of surli/repairnator:latest and now you're using the new docker image!

for ```dockerpool``` to properly test if my changes have worked?

surli:
> Okay, I think I am getting slightly confused now.

I can imagine that ;)
To be honest I suggested you to take this issue quickly in order to have that discussion about Repairnator architecture :) 

So let's go back to this architecture.
All Repairnator scripts that are using the pipeline, are using a docker container. This docker container is created from a docker image of repairnator.
For example when you call `docker run hello-world` there is a docker image called `hello-world` and a docker container is automatically created and launched from that image. 

Now two questions remains: 
  1. Which image is used to create/run the containers
  2. How to create the image

The first one is easy: we use the information placed in `repairnator.cfg`. 
For the second one, we normally use the Dockerfile placed in `docker-images/pipeline-dockerimage`. This docker file calls a script which is https://github.com/Spirals-Team/repairnator/blob/master/repairnator/docker-images/pipeline-dockerimage/build_repairnator.sh: this script will only download a jar from Maven based on a version placed on `version.ini` file. 

Now, when you're developing new features, you cannot rely on this jar available on Maven, since it has not been deployed yet. So that's why you have to build your own docker image: I removed in `docker-images/pipeline-dockerimage-dev` the call to the `build_repairnator.sh` script, and instead you have to manually put the proper repairnator-pipeline.jar.

So to answer your questions:
>  Is skipDelete default behaviour in repairnator? Or at least for pipeline?

The default behaviour is to delete the container after its run in dockerpool. That's why we want an argument to allow skipping the deletion.

> should the changes in docker-images/pipeline-dockerimage remain

Actually no because the pipeline doesn't need the argument.

btellstrom:
Okay, I think I understand what you are saying. Does that mean that for this case I do not need to construct a new repairnator-pipeline.jar since there are no changes to it? And it will always display skipDelete=false in logs from ```repairnator-pipeline``` just because it doesn't need it and it is an effect from the ```toString()``` method from ```RepairnatorConfig``` (this is under the assumption that we leave repairnator-pipeline in it's current state)?

And once I have removed the changes to ```pipeline-repairnator```, should I have to do anything else to finish this PR? Assuming I have understood you correctly I am slightly lost at how I would confirm the option other than running the rt_scanner for a time (10minutes didn't seem to cut it).

surli:
>  Does that mean that for this case I do not need to construct a new repairnator-pipeline.jar since there are no changes to it? And it will always display skipDelete=false in logs from repairnator-pipeline just because it doesn't need it and it is an effect from the toString() method from RepairnatorConfig (this is under the assumption that we leave repairnator-pipeline in it's current state)?

Yes and yes. :)

> And once I have removed the changes to pipeline-repairnator, should I have to do anything else to finish this PR?

I don't think so, no. I will review the PR afterwards.

> I am slightly lost at how I would confirm the option other than running the rt_scanner for a time (10minutes didn't seem to cut it)

Actually there should be a better way to do it: I created a project called https://github.com/surli/failingProject which is purposely failing on Travis and which should be repairable by Repairnator. This project ran everyday on Travis with a cron. 
Then you can call launch_repairnator.sh on a list containing only `surli/failingProject` to use repairnator with the standard scanner (and not the rt_scanner) on a given period of time: you have first to change the repairnator.cfg to specify the date range. Then you specified something like a week, it should run 7 different docker containers corresponding to the failing builds.

repairnator/scripts/config/repairnator.cfg
```diff
@@ -27,6 +27,7 @@ SCANNER_NB_HOURS=1 # Number of hours to inspect to get last builds (e.g. 4 will
 
 NB_THREADS=4 # Number of concurrent docker container to run
 DAY_TIMEOUT=1 # Global timeout to stop the docker execution
+SKIP_DELETE=1 # Skip the deletion of docker containers
```
surli: IMHO the option should be placed at the end of the file under `### Switches` category. And its default value should be 0. Indicate that 1 would mean to keep the docker containers instead of deleting it.

repairnator/scripts/repair_buggy_build.sh
```diff
@@ -49,13 +49,16 @@ echo "Pull the docker machine (name: $DOCKER_TAG)..."
...
 LOG_FILENAME="repairnator-pipeline_`date \"+%Y-%m-%d_%H%M\"`_$BUILD_ID"
+
 DOCKER_ARGS="--env REPAIR_MODE=repair"
 DOCKER_ARGS="$DOCKER_ARGS --env BUILD_ID=$BUILD_ID"
 DOCKER_ARGS="$DOCKER_ARGS --env GITHUB_OAUTH=$GITHUB_OAUTH"
 DOCKER_ARGS="$DOCKER_ARGS --env LOG_FILENAME=$LOG_FILENAME"
 DOCKER_ARGS="$DOCKER_ARGS --env RUN_ID=$RUN_ID"
 DOCKER_ARGS="$DOCKER_ARGS --env OUTPUT=/var/log"
 DOCKER_ARGS="$DOCKER_ARGS --env REPAIR_TOOLS=$REPAIR_TOOLS"
+DOCKER_ARGS="$DOCKER_ARGS --env SKIP_DELETE=$SKIP_DELETE"
```
surli: So you can remove this one.

repairnator/scripts/config/repairnator.cfg
btellstrom: Fixed.

repairnator/scripts/repair_buggy_build.sh
btellstrom: Fixed.

btellstrom:
I was attempting to run the ```launch_repairnator.sh``` with a file containing only ```surli/failingProject```, however that failed. It needed a row before ```surli/failingProject```. 

Also, when I added a row before it said it only found 1 build ( I tried with the 2 configurations 
```
LOOK_FROM_DATE=22/06/2018 # Use with the following one, when wanting to scan a period of time (format: dd/MM/yyyy)
LOOK_TO_DATE=29/06/2018
```
and 
```
SCANNER_NB_HOURS=161
```
)
Adding extra lines to the list resulted in as many more ```builds found``` as lines added. Have I misunderstood what you meant by a list? I assumed the end of a link on the form https://travis-ci.org/**surli/failingProject** and in a regular file.

Also, thank you for the help :)

surli:
> I was attempting to run the launch_repairnator.sh with a file containing only surli/failingProject, however that failed.

OK that was not what I meant. 
`launch_repairnator.sh` works with 0 or 1 argument: when the argument is given, it must be a list of build IDs. In that case we assume that you already computed the list of IDs on which to run repairnator.

Here it's not the case: you the list of IDs is not available yet.
In repairnator.cfg there is an argument which is:
`REPAIR_PROJECT_LIST_PATH=$HOME_REPAIR/project_list.txt`. This is the actual list of project to consider for the scanning process. It is then in this file that you should put `surli/failingProject`. 

Now a side not: in the pipeline we use a list of ignored projects: `surli/failingProject` is on that list by default. The reason is that we do not want our on-production bot to spam us with fixes of surli/repairnator.
It means that if you need to run repairnator-pipeline on `surli/failingProject` you need first to update that list. Now here it shouldn't be a problem, because you don't really need to have a full run of  the pipeline for your test: you just need the dockerpool to call the docker containers with the build ids.

btellstrom:
Running it on my laptop crashed the laptop, but the containers are not deleted when running ```launch_repairnator.sh``` with ```SKIP_DELETE=1``` and are deleted when it is 0 instead when I ran it on my desktop. In other words, it seems to be working at least.

surli:
Don't forget to edit the title from `WIP` to ready when it is ;)
Sorry for the tedious process to achieve this one, but now you have a lot of information about how repairnator works!

monperrus:
great!

Issue #665, opened 2018-07-16
All final steps should be skipped in the case of pipeline being stopped without bug reproduction


fermadeiral:
Log used as ref: repairnator-pipeline_2018-07-16_0429_404269638_1a321fce-b0e8-4450-af34-5c48274cc198.log

For Repairnator, if the bug was not reproduced, the final steps should be skipped. It may save time.

For Bears, if the bug and/or the patch were not reproduced, the final steps should also be skipped (I think this is already done).

PR #673, opened 2018-07-17
Skip final steps in the case of pipeline being stopped without bug reproduction

fermadeiral:
This PR targets issue #665

### DIFF
--- a/repairnator/repairnator-pipeline/src/test/java/fr/inria/spirals/repairnator/process/inspectors/TestProjectInspector.java
+++ b/repairnator/repairnator-pipeline/src/test/java/fr/inria/spirals/repairnator/process/inspectors/TestProjectInspector.java
@@ -171,8 +171,6 @@ public class TestProjectInspector {
 
         Map<Class<? extends AbstractStep>, StepStatus.StatusKind> expectedStatuses = new HashMap<>();
         expectedStatuses.put(BuildProject.class, StepStatus.StatusKind.FAILURE); // step supposed to fail by this test case
-        expectedStatuses.put(CommitProcessEnd.class, StepStatus.StatusKind.SKIPPED); // the repo won't be initialized, thus there will be an error on this step so it will be skipped
-        expectedStatuses.put(PushProcessEnd.class, StepStatus.StatusKind.SKIPPED); // no remote info provided
 
         this.checkStepStatus(stepStatusList, expectedStatuses);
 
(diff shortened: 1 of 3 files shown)


repairnator/repairnator-pipeline/src/test/java/fr/inria/spirals/repairnator/process/inspectors/TestProjectInspector.java
```diff
@@ -171,8 +171,6 @@ public void testFailingProjectNotBuildable() throws IOException {
 
         Map<Class<? extends AbstractStep>, StepStatus.StatusKind> expectedStatuses = new HashMap<>();
         expectedStatuses.put(BuildProject.class, StepStatus.StatusKind.FAILURE); // step supposed to fail by this test case
-        expectedStatuses.put(CommitProcessEnd.class, StepStatus.StatusKind.SKIPPED); // the repo won't be initialized, thus there will be an error on this step so it will be skipped
```
surli: Could you explain why you removed those? I'm not sure to understand.

fermadeiral: Yes.
It's because in the past, when the pipeline failed without bug reproduction for Repairnator (or bug/patch for Bears), the execution flow passed throught these two steps (which are final steps), which were skipped from inside them when there is no push info (which we don't provide in the tests, and this is right). Thus, in the `expectedStatuses` variable, there was info on those steps, i.e. that they were skipped by lack of push info.
However, with the changes in this PR, the final steps are always directly skipped in `AbstractStep` if the bug was not reproduced for Repairnator (or bug/patch were not reproduced for Bears). So the execution flow does not pass throught the final steps when the pipeline is broken, and for that reason the variable `expectedStatuses` will not contain status from the final steps.
The test I changed is supposed to result in a broken pipeline, since it's a not buildable build, so the final steps should not be executed.

fermadeiral: The need to change this test is actually a good sign that the changes in this PR are working, i.e., no bug reproduction, no final steps' execution.

surli: ok thx

Issue #546, opened 2018-05-11
Improve mail notification for end of process


surli:
Add the following information in the sent email:
 - when the repair attempt started
 - when the repair attempt ended
 - Repairnator process ID PID
 - Repairnator uptime
 - on which machine

And some feedback about what it did maybe.

monperrus:
@surli: are emails sent through SMTP or through some web services?

PR #678, opened 2018-07-31
Ready: Improve mail notification for end of process

btellstrom:
Attempt at resolving issue #546. 

If I have understood the end of process notification correctly, then the start and end of the repair attempt was already mentioned in the mail. The only thing lacking was the duration, or if duration is not the same as uptime I have made a mistake. Should I perhaps `config.getInstance().getDuration().toString()` instead?


surli:
No your solution looks good to me. I wait for the CI and I'll merge it.

btellstrom:
> And some feedback about what it did maybe.
Nothing needed for that?

surli:
> Nothing needed for that?

I did not specify it, and I don't feel the need anymore so let's move on

PR #684, opened 2018-08-02
Add some tests for jruby/jruby

### LINKED ISSUE #652: test: test ComputeSourceDir / ComputeTestDir with jruby/jruby
The bug #650 was obtained when trying to fix a build from https://github.com/jruby/jruby. It looks like a project using heavily multi-modules.

We should test ComputeSourceDir and ComputeTestDir with their pom.xml file.

surli:
fix #652


surli:
@fermadeiral @btellstrom I took the same pom.xml than those from jruby/jruby some I'm not checking really how it works with the properties to setup another source folder, but I think it's already checked in another test.

Then I check in an exhaustive way how it works from one module to its parent to another module to check the source and test directories. Tell me what you think about it.

repairnator/repairnator-pipeline/src/test/java/fr/inria/spirals/repairnator/process/step/paths/TestComputeSourceDir.java
```diff
@@ -28,8 +28,12 @@
 import java.util.Optional;
 
 import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
```
fermadeiral: These imports are not used.

fermadeiral:
Looks good to me. But we still miss tests on the `allModules` field. Maybe we could open a specific issue for that.

surli:
>  But we still miss tests on the allModules field

My bad. I'll improve my tests

btellstrom:
Looks good to me to, appreciate the comments as well.

surli:
I just added the missing tests for allModules: there was indeed a bug with this one.
When calling the method with allModules from a module, it was getting the source (or test) dir from its parent, but not from the module of its parent. 
Now it's fixed.

surli:
it's ready for merging @fermadeiral

fermadeiral:
Oh, ok. I didn't know that I was supposed to merge.
Thank you @surli for the tests.

PR #696, opened 2018-08-03
Add checks for json files in checkbranches module (continuation)

fermadeiral:


fermadeiral:
Docker images for checkbranches are now working.

Issue #476, opened 2018-03-23
repairnator.json: clarify dates in it


surli:
The dates of the beginning of the process and the creation of patches are missing. They should be included.

fermadeiral:
I would like to add to this issue the refactoring of `repairnator.json`. This includes:

1) the refactoring of `repairnator.json`'s schema based on the one we created together for `bears.json`, including the addition of properties in the schema for repairnator (like the dates mentioned in this issue);
2) the implementation of the collection of those properties;
3) the creation of a test for it (based on the one we have started in `TestMetrics4BearsJsonFile`).

surli:
In order to do this refactoring properly we should do the following steps:
 1. Create a new schema for repairnator that satisfy us,
 2. Try to write an automatic translation from the old schema to the new one, maybe using this tool: https://github.com/dvdln/jsonpath-object-transform

fermadeiral:
Given a schema for `repairnator.json`, I can be responsible for working on the implementation on the json file generation on the pipeline module since that will be based on the one I created for Bears (which will be mostly cleaning, and differing specific stuff between Bears and Repairnator), I can update the test in `TestMetrics4BearsJsonFile.testBearsJsonFileWithFailingBuild` (which does not work properly by now), and I can setup the tests for Repairnator on the check branches properly (what I added now in #697 is on the old json file schema, so I would need to create new testing branches).

PR #702, opened 2018-08-05
Rename things related to bears.json file

### LINKED ISSUE #615: Review word "metrics" being used everywhere.
We use the word "metrics" to refer to a set of properties, which includes quantitative and qualitative data. In rare cases, we use the word "properties". I think the latter is better to express what we are referring to, because when I read "metrics", I think on numbers only. For instance, the classes `Metrics` and `Metrics4Bears` have a lot of "info" not related to number. I propose the usage of the word "properties" everywhere. WDYT?
Despite the fact that performing a solution for this issue will affect several different locations in the project, I think this is a good-first-issue.

fermadeiral:
This PR does not close any issue, but it is a start towards the new `repairnator.json` generation (related to #476), which also contributes to close #615.

### DIFF
--- a/repairnator/repairnator-core/src/main/java/fr/inria/spirals/repairnator/serializer/SerializerType.java
+++ b/repairnator/repairnator-core/src/main/java/fr/inria/spirals/repairnator/serializer/SerializerType.java
@@ -22,7 +22,7 @@ public enum SerializerType {
     SCANNER4BEARS("scanner4bears"),
     HARDWARE_INFO("hardwareInfo"),
     METRICS("metrics"),
-    METRICS4BEARS("metrics4bears"),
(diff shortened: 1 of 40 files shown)


fermadeiral:
The build is broken, and I still can't really understand why, so I will leave this PR as WIP until I find a way to fix it.

fermadeiral:
@surli, could you check this PR for me, please, to try to find the issue? I will close this PR if I can't fix that today, because I can't find the problem with it.

surli:
Sorry I did not check this one. I might have a look tomorrow.

fermadeiral:
Don't worry about this one. I will create a new PR and I will do a commit for each refactoring to find what is making the build failing.

PR #710, opened 2018-08-06
Rename things related to bears.json file

fermadeiral:
The code elements related to the generation of the `bears.json` are usually named with "4bears". However, those elements will be shared between `bears.json` and `repairnator.json` generation (related to issue #476), thus the names need to be generic. This PR renames the mentioned code elements. Additionally, since the renaming will already happen here, this PR also contributes towards to close the issue #615.


fermadeiral:
@surli, I have here the same issue than in the PR I closed yesterday. I can't find the problem. I'm kind of losing my mind here with this problem. As I tested here, the test fails when I change the name of its folder. So I'm going to do the refactorings I need but I will not touch in the test, this will stay with the old name.

surli:
And does it fail on your machine? Just run the test on my machine without any problem.

fermadeiral:
Exactly. In my machine the test also passes. I really don't have idea why this is failing on Travis.

surli:
Actually I just manage to reproduce the bug by launching the entire test suite. So it seems related to the interaction between tests.

fermadeiral:
This is so weird for me. I don't understand how renaming the folder is causing the failure.

Ok, so let me propose you something: I do all the renaming I need, and I let the renaming of the tests out of this; then you review/merge this PR; then I will try to fix the issue with the test in another PR. WDYT? Because a lot of things need to be renamed, and I'm stuck with this problem since sunday.

surli:
It's ok for me. Actually I'm even in favor of not renaming this one for now: there're more urgent stuff ;)

fermadeiral:
> [...] there're more urgent stuff ;)

Can I do something to help?

surli:
> Can I do something to help?

You already help :) But yeah there's the issue with repairnator.json if you're interested in starting it...

fermadeiral:
> But yeah there's the issue with repairnator.json if you're interested in starting it...

In fact I'm really interested in closing this issue because it will clean and solve a lot of unfinished parts in the code. I will propose a PR soon to start to work on that.

fermadeiral:
@surli I believe this one is ready.

PR #711, opened 2018-08-07
Refactor repairnator.json schema

fermadeiral:
This PR is the first step to address the issue #476.


fermadeiral:
@surli, ok, I have identified the properties in the old schema that we don't have in the new schema. Let's start to discuss them so I can change the new schema. First, I would like to start to talk about the following properties:

- hostname
- metrics.NbCPU
- metrics.FreeMemory
- metrics.TotalMemory

Do you want to keep all those properties? If yes, what do you think on putting those 4 properties together in an element called `machineInfo`, and this element would be inside the `reproductionBuggyBuild` element that we already have in the schema?

surli:
First, yes I want to keep those info.
Second, ok to put them in `machineInfo` inside `reproductionBuggyBuild`

fermadeiral:
Moving on...now I would like to discuss the property object `metrics.FreeMemoryByStep`. Would you like to keep this one? If yes, I still don't know how to do it. Just to remember, we refactor `metrics.StepsDurationsInSeconds`, which is a cousin of `metrics.FreeMemoryByStep`, for `processDurations` where we have `cloning`, `building`, `testing`, and `fixing`, but they don't include ALL the executed steps: in that time, we selected the ones that we wanted to record the duration.

surli:
Let's remove that one. We have it in the serialization it's enough.

fermadeiral:
I believe this is the last one that we have in the old but not in the new that we should discuss: `metrics.AngelicValuesByTest`. If I properly understood, this is on Nopol, right? What do you think about doing with it?

Moreover, let me start to discuss what you first wrote in #476: "The dates of the beginning of the process and the creation of patches are missing. They should be included.". What do you think at creating something like this:

The date of the beginning of the process we already have in the new schema: it's `reproductionBuggyBuild.reproductionDateBegin`. So this one is already ok, right?

About the date of the creation of the patches I'm a little but confused. What exactly should this be? Because there are several tools, so let's supposed that 2 tools generated at least one patch. Would you like to record two dates, one for each first patch by a tool?

fermadeiral:
@surli so the schema is ready for your review. I will update later (in another PR) the example and the mapping files. I just need the schema by now to be able to do the properly generation and to show you a resulted branch.

resources/repairnator-schema-new.json
```diff
@@ -975,6 +982,54 @@
             "Apr 5, 2018 6:12:38 AM CET"
           ]
         },
+        "machineInfo": {
```
surli: machineInfo should be a required property, no?

fermadeiral: Done.

fermadeiral:
> I think you forget to edit the name of the schema at the beginning

Done.

fermadeiral:
@surli, you said "Let's remove that one. We have it in the serialization it's enough." when I asked on `metrics.FreeMemoryByStep`. Where do we have it in the serialization? I don't find it.

fermadeiral:
> It should be saved in the metrics serializer. It's kind of hidden because I get the whole Metrics object.

I see. I will need to touch on that serialization too then, to keep that, but I don't see any problem, I think it will be ok.

So for me the schema is ready.

surli:
So, let's merge it then!

PR #713, opened 2018-08-07
Fix fixme in CloneRepository

fermadeiral:
Do you think this solve the fixme?

### DIFF
--- a/repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/step/AbstractStep.java
+++ b/repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/step/AbstractStep.java
@@ -267,7 +267,7 @@ public abstract class AbstractStep {
             }
         }
 
-        if (this.config.isClean()) {
+        if (this.inspector.getRepoLocalPath() != null && this.config.isClean()) {
             try {
                 FileUtils.deleteDirectory(this.inspector.getRepoLocalPath());
             } catch (IOException e) {
--- a/repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/step/AbstractStep.java
+++ b/repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/step/AbstractStep.java
@@ -257,7 +257,7 @@ public abstract class AbstractStep {
         return this.inspector.getJobStatus().getPomDirPath() + File.separator + Utils.POM_FILE;
     }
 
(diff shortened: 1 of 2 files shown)


repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/step/AbstractStep.java
```diff
@@ -267,7 +267,7 @@ protected void cleanMavenArtifacts() {
             }
         }
 
-        if (this.config.isClean()) {
+        if (this.inspector.getRepoLocalPath() != null && this.config.isClean()) {
```
surli: Actually the `getRepoLocalPath` is not null: if you look line 13 of `CloneRepository` it shows that we use this value.
However we could check the existence of `getRepoLocalPath() + ".git"`.

fermadeiral: Hmm, I see. Check out the new commit and tell me what you think, please.

fermadeiral: After I commit that I got doubt about something: does the check you told me is supposed to wrap the entire body of the method `cleanMavenArtifacts`? Because if yes, I need to change again.

repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/step/AbstractStep.java
```diff
@@ -267,7 +267,8 @@ protected void cleanMavenArtifacts() {
             }
         }
 
-        if (this.inspector.getRepoLocalPath() != null && this.config.isClean()) {
+        File repoDir = new File(this.inspector.getRepoLocalPath() + File.separator + Utils.REMOTE_REPO_EXT);
```
surli: Prefer using `new File(children, repo);` here it's `new File(this.inspector.getRepoLocalPath(), Utils.REMOTE_REPO_EXT)`

fermadeiral: Done.

fermadeiral:
Simon, the method `cleanMavenArtifacts` looks like actually a clean on maven and on the repo. Maybe we should rename it?

surli:
We can rename it yes.

fermadeiral:
WDYT on the new name?

PR #715, opened 2018-08-08
Refactor repairnator.json generation

fermadeiral:
This PR adds the implementation for the generation of `repairnator.json` according to the new schema (PR #711). As a result, the classes `Metrics` and `MetricsSerializerAdapter` will be deleted, and also the method `JobStatus.writeProperty`.


fermadeiral:
@surli, I have test failures here because of the property `projectMetrics.numberLines`. The result from cloc is 0. I don't know how to fix that. In the last case, I will put this property as to be ignored in the test.

surli:
Ok for ignoring the property, put a comment why you do that in the test then.

fermadeiral:
@surli, the info on the machine is recorded only when the pipeline is ending. However, the freeMemory should not be calculated when it's starting? For each step, it's the same thing: the freeMemory is recorded after the execution, but it should not be before?

surli:
> For each step, it's the same thing: the freeMemory is recorded after the execution, but it should not be before?

Not necessarily as long as we keep the same behaviour everywhere.

fermadeiral:
@surli, this PR seems just fine for me, you can start review it if you want, but the build should break because of the check branches. I will now generate the testing branches again.

fermadeiral:
@surli, the testing branches are updated. Now it's really to wait for Travis. If you would like to have a look in the new `repairnator.json`, you can find an example here: https://github.com/fermadeiral/test-check-branches/tree/surli-failingProject-208897371-20170308-060702

fermadeiral:
@surli, I have no idea why the last build is failing.

surli:
It's again related with the previous bug. So let's ignore that test with a strong comment for future fixing.

surli:
Once this PR is merged, I start the release process. It might take time so I prefer to begin it early.

fermadeiral:
Will it be still possible for me to change the commit messages while you start the release process?

surli:
> Will it be still possible for me to change the commit messages while you start the release process?

Start a new PR by creating a new branch from this one. A do the change on the commit msg now.

PR #729, opened 2018-08-28
Remove bad test (the build's commit can't be reached anymore due to branch deletion)

### LINKED ISSUE #728: Fix TestComputeSourceDir.testComputeSourceDirWithMultiModuleProject5
The build in PR #727, which can't be failing by its own changes, is failing due `TestComputeSourceDir.testComputeSourceDirWithMultiModuleProject5`. Such test uses a build from the project jacoco/jacoco, and apparently the problem happens when trying to do the checkout of the build's commit: I believe the issue is due the deletion of the branch where the commit was done (see https://github.com/jacoco/jacoco/pull/496), which might mean that any build triggered on repairnator will fail starting from now.

IMHO, we should remove this test from repairnator: 1) it's buggy, we can't resolve the commit; 2) there are already 4 other tests of this same type (i.e. that test the step `ComputeSourceDir` on multi-module project), and 3) the main assertion in this test is even comment.

fermadeiral:
Fix #728

### DIFF
--- a/repairnator/repairnator-core/src/main/java/fr/inria/spirals/repairnator/Utils.java
+++ b/repairnator/repairnator-core/src/main/java/fr/inria/spirals/repairnator/Utils.java
@@ -74,6 +74,10 @@ public class Utils {
         return "http://github.com/"+repoSlug+"/commit/"+commitId;
     }
 
+    public static String getBranchUrl(String branchName, String repoSlug) {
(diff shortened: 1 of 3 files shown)


fermadeiral:
@btellstrom could you review this PR for me? If everything is ok for you, could you merge this PR in the master branch?

PR #731, opened 2018-08-30
Check if container exit code is really 0 before serializing it as TREATED

fermadeiral:
When executing dockerpool, I killed a container and for my surprise, it was serialized as TREATED, which should indicate that the execution of an instance from `RunnablePipelineContainer` was just fine, but in the dockerpool log is possible to see that me killing the container resulted in an exit code different from 0. This PR intends to serialize a `RunnablePipelineContainer` as ERROR instead of TREATED when the container exit code is not 0.


fermadeiral:
> This seems odd, should the killing of the container not result in a Dockerexception, that would then get caught by the catch-block underneath (row 159)?

I thought the same thing as you. I don't really understand why no catch was reached since the docker container indeed returned a non-zero exit code. Well, maybe we are missing something, so this is why we think this is strange. In any case, this PR makes sure that a serialization will be TREATED only with exit code 0.

btellstrom:
Okay, at least I'm not the only one confused :)

I'll merge it then.

PR #738, opened 2018-10-04
Fix scanner for Bears

fermadeiral:
The scanner for Bears is bringing a lot of pairs of builds that should not enter for reproduction:

1) when we have a pair of failed and passed builds, it should be checked if the failed build failed in Travis due to test failure;
2) when we have a pair of passed and passed builds, it should be checked if there is one java file between the two builds that is not test file.

This PR fixes that.


fermadeiral:
> I am however curious about one thing, where the old builds used for the test not sufficient for the new test, or did you change them for some other reason?

Yes, the old pair of builds are sufficient for the new negative test. I had chosen another pair for the new negative test without realizing in fact, because I had a list here of pairs that the scanner should not get.
I put the old pair of builds in the new negative test just now :-)

btellstrom:
Okay, then I have no more questions :)

PR #746, opened 2018-10-12
Automatic Repair of Checkstyle Errors

bloriot97:


fermadeiral:
Hi @bloriot97. I would like to review this PR when it's ready since changes here have an impact on both Bears and Repairnator.

surli:
Hi @bloriot97 thanks for this contribution! Be careful though, it seems you accidentally committed some .class files here.

monperrus:
It's probably not the final version of this, but it passes all the tests, so we can merge it. (Repairnator is still in highly evolvable mode).

@bloriot97 will build upon this in the next days.

If something is broken that was not tested, we will add the missing test in a future PR.

monperrus:
Thank you @bloriot97

PR #749, opened 2018-10-17
fix security issue

monperrus:
we want the warning banner to disappear before noon :)


monperrus:
Flute, it seems like 2.9.5 is not backward compatible.

Could you have a look?

PR #751, opened 2018-10-17
Website enhancements (vue, api helper, notifications, buildinfo modal) 

bloriot97:
Sorry for the mess...

So in this single pr/commit there are many different improvements.
- api helper to simplify requests to the api
- very basic handling of notifications 
- integration of vue.js 
- implementation of a modal to have more information about the build reproduction 
- update of the doc

If you want to have a sight on the [website]( http://51.38.235.157/).


website/repairnator-site/app/scripts/api.js
```diff
@@ -0,0 +1,13 @@
+function getApiUri(path){
+  const apiHost = 'http://51.38.235.157:4040/repairnator-mongo-api';
```
surli: I know I didn't do it either with the first version, but since you are refactoring it, couldn't you use [dotenv](https://www.npmjs.com/package/dotenv) to put the domain and port in an environment file used for deployment? It would be cleaner I think :)

bloriot97: Good idea.
But as the website is is built by gulp how should I implement it ?
Because i won't be able to require dotenv from the webpage. Then I guess I should implement it in the gulpfile to put it in the js ?

surli: https://www.freshconsulting.com/how-to-organize-your-gulp-js-development-builds-for-multiple-environments/

bloriot97: Thanks :)

bloriot97:
On @surli advice I also added config options for deployment using config. 
I used config instead of dotenv because I think it is more convenient to manage different environments.

website/repairnator-site/README.md
```diff
@@ -17,6 +17,9 @@ $ npm install -g gulp bower
 $ npm install
 $ bower install
 
+# copy the config files (and edit them)
```
surli: why do you need multiple config files?

bloriot97: So that you can specify different variables depending on your environment. 
Config handle different environments by loading the specific file and the default option file.

surli: why not writing the proper config file to deploy based on some objects contained in only one conf file? 

something like:

```json
{
  "constants": {
    "default": {
      "host": "http://defaulthost",
      "port": "4444"
    },
    "dev": {
      "host": "http:localhost",
      "port": "88"
    },
    "prod": {
      "host": "http://fancyurl",
      "port": "80"
    }
}
```

then you have only one file to rely one for all config.

bloriot97: It is also an option. But then I would not be using [config](https://www.npmjs.com/package/config) anymore. 
But it is no problem I can also create a small config loader to handle this.

surli: hm indeed I didn't check that you were using this lib. Your call then: it's really a detail here. I can merge the PR like this.

bloriot97: I think both solutions are good so your can merge like that.

PR #757, opened 2018-10-22
added options to the config-file 

btellstrom:
In the previous smtp PR (#736) I forgot to add the options to the configuration file.


monperrus:
@bloriot97 do you review and merge?

btellstrom:
It is now able to successfully send emails about patches.

fermadeiral:
> In the previous smtp PR I forgot to add the options to the configuration file.

@btellstrom could you please tell me which is the previous PR (the number I mean)? I would like to take a look on what was done before to better understand this one.

btellstrom:
It was PR #736

repairnator/repairnator-core/src/main/java/fr/inria/spirals/repairnator/notifier/engines/EmailNotifierEngine.java
```diff
@@ -46,7 +48,14 @@ protected PasswordAuthentication getPasswordAuthentication() {
         }
 
         try {
-            this.from = new InternetAddress(smtpUsername);
+            if(!smtpUsername.contains("@")) {
+                /* apparently kths mailserver does not accept complete mailaddresses, but sender needs one. This assumes that 
+                 * the host name after "smtp." is the actual internet address. Might need modifications.
+                 */
+                this.from = new InternetAddress(smtpUsername + "@" + this.properties.getProperty("mail.smtp.host").substring(5));
```
fermadeiral: I don't understand what happens around `.substring(5)`. Could you explain it to me?

fermadeiral:
It seems good in general to me, but I would like to know if there is some specific reason for changing only one docker image instead of the three.

repairnator/repairnator-core/src/main/java/fr/inria/spirals/repairnator/notifier/engines/EmailNotifierEngine.java
fermadeiral: My concern with hard-coded things like that is that it's too abstract, and other contributors might not understand why the code is this way (I don't so far, for instance). Maybe we should even think on encapsulating this in some method (like in the `Utils` class we have) and creating like two test cases for it.

btellstrom: KTHs smtp-server does not accept logins with "@kth.se" specified, but the sender field needs to contain "@kth.se". I have here assumed that most smtp-servers have "smtp." as the first part of their name, which happens to be 5 characters long. 

A more elegant solution would perhaps be to have the fully specified name, and if it can not connect using that to try to connect using only that before the "@"?

btellstrom:
> It seems good in general to me, but I would like to know if there is some specific reason for changing only one docker image instead of the three.

repairnator/repairnator-core/src/main/java/fr/inria/spirals/repairnator/notifier/engines/EmailNotifierEngine.java
btellstrom: I am not sure I understand what you mean by encapsulating it in some method. Do you mean we should add to `Utils` a method which converts a not completely specified email-address to a completely specified address by the same method as above, and then write some tests for it?

fermadeiral: > KTHs smtp-server does not accept logins with "@kth.se" specified, but the sender field needs to contain "@kth.se". I have here assumed that most smtp-servers have "smtp." as the first part of their name, which happens to be 5 characters long, followed by the actual domain name.

Let me see if I get this right: (when KTH...) the `smtpUsername` needs to be full (containing the two sides of "@") to instantiate `InternetAddress`, but not for instantiating `PasswordAuthentication`?

> A more elegant solution would perhaps be to have the fully specified name, and if it can not connect using that to try to connect using only that before the "@"?

Maybe, but first I need to completely understand the issue that you are facing.

repairnator/docker-images/bears-dockerimage/pipeline_launcher4Bears.sh
```diff
@@ -26,6 +26,10 @@ export MONGODB_HOST=
 export MONGODB_NAME=
 export PUSH_URL=
 export SMTP_SERVER=
+export SMTP_PORT=
```
fermadeiral: The `args` above needs to be updated.

repairnator/docker-images/pipeline-dockerimage/pipeline_launcher.sh
```diff
@@ -30,6 +30,10 @@ export MONGODB_HOST=
 export MONGODB_NAME=
 export PUSH_URL=
 export SMTP_SERVER=
+export SMTP_PORT=
```
fermadeiral: The `args` above needs to be updated.

repairnator/repairnator-core/src/main/java/fr/inria/spirals/repairnator/notifier/engines/EmailNotifierEngine.java
fermadeiral: > I am not sure I understand what you mean by encapsulating it in some method. Do you mean we should add to Utils a method which converts a not completely specified email-address to a completely specified address by the same method as above, and then write some tests for it?

Yes. It would be a method with like one/two lines. But this could be good for testing. For instance, we could create 2 test cases, one that something needs to be changed in a given `smtpUsername` (we could test it with yours, for instance), and another one that no change is needed.

btellstrom: >Let me see if I get this right: (when KTH...) the `smtpUsername` needs to be full (containing the two sides of "@") to instantiate InternetAddress, but not for instantiating `PasswordAuthentication`?

Basically. When authenticating, one must only provide that to the left of "@", but when actually sending the message the `from` part needs both sides of the "@".

fermadeiral: > Basically. When authenticating, one must only provide that to the left of "@", but when actually sending the message the from part needs both sides of the "@".

I see what you mean then =/

I think we could let this how you did now, and in the future when we find some case that this will not work, we think in another solution. WDYT?

For now, I was thinking that something like `this.properties.getProperty("mail.smtp.host").replace("smtp.","")` might be easier to understand than `this.properties.getProperty("mail.smtp.host").substring(5)`. WDYT?

btellstrom: >For now, I was thinking that something like `this.properties.getProperty("mail.smtp.host").replace("smtp.","")` might be easier to understand than `this.properties.getProperty("mail.smtp.host").substring(5)`. WDYT?

Yes, I agree.

repairnator/docker-images/bears-dockerimage/pipeline_launcher4Bears.sh
btellstrom: Done.

repairnator/docker-images/pipeline-dockerimage/pipeline_launcher.sh
btellstrom: Done.

fermadeiral:
For me, this PR is ready to be merged.

bloriot97:
For me everything is fine, I will merge it once Travis done.

btellstrom:
Uh, apparently the NPEfix fails on this [build](https://travis-ci.org/surli/failingProject/builds/252712792) now. I am uncertain as to why...

bloriot97:
Merged

PR #766, opened 2018-10-24
[website] Put the components into a proper component folder

bloriot97:
I put the vue components into a component folder.
I will make both the html and the js more readable. 

All the components will be added in a component.js file that will be loaded before the scripts.

### DIFF
--- a/website/repairnator-site/app/index.html
+++ b/website/repairnator-site/app/index.html
@@ -63,51 +63,6 @@
       </div>
     </section>
 
-    <script type="text/x-template" id="pagination-template">
-      <nav aria-label="Page navigation" style="padding: 0px 30px">
-        <ul class="pagination">
-          <li class="page-item" v-bind:class="{disabled: !page}">
-            <a class="page-link" @click="prevPage" aria-label="Previous">
-              <span aria-hidden="true">&laquo;</span>
-              <span class="sr-only">Previous</span>
-            </a>
-          </li>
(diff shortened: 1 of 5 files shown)


website/repairnator-site/app/index.html
```diff
@@ -63,51 +63,6 @@ <h2 class="promo__title">What is Repairnator doing right now?</h2>
...
-          <tbody>
-            <tr v-for="entry in data" v-bind:class="entry.status">
-              <td v-for="key in columns">
-                <span v-html="entry.data[key.id]"> </span>
-              </td>
-            </tr>
-          </tbody>
-        </table>
-      </div>
-    </script>
```
surli: So basically you removed this whole block to put it in a template in a separate .js to be used as a component. Just a question though: how the templates are injected in the right place in the html? I just see you injected the .js.

bloriot97: The templates are no longer injected in the html, the component will use a text template as [here](https://github.com/Spirals-Team/repairnator/pull/766/files/50b5636bb6212b8600cf8370bfb2ec651fe0e648#diff-58fc2b47b57fcd35a3db064b1b64dcf1R26) for example.
And then the component will still be inserted at the right place [here](https://github.com/Spirals-Team/repairnator/pull/766/files/50b5636bb6212b8600cf8370bfb2ec651fe0e648#diff-3f4e211be47623287ac1b62e8534ce3dR62) just the same.
I also thought putting the templates in the html with gulp but this is a bit messy and both are rendered in the brother then in terms of performance it is the same. 

I think putting everything in a single js file is more clear though.

PR #778, opened 2018-10-29
Refactoring on dates.

### LINKED ISSUE #773: Refactoring on dates.
@surli and @bloriot97, from the issue #772, I realized that we always do `new Date()` to refer to the end of the reproduction, but this is prone to inconsistencies (very difficult to happen, but it could happen) in the db, thus we might refactor that to always record the same exact date, which would be in `ReproductionBuggyBuild.getReproductionDateEnd` I think (to be precisely checked), WDYT?

fermadeiral:
This PR intends to fix #773.


monperrus:
ready?

monperrus:
Thanks a lot!

Issue #767, opened 2018-10-25
repair_buggy_build.sh script creates a pull-request even if CREATE_PR is set to 0


btellstrom:
We do not want Repairnator to create pull request without explicitly stating so. If this also happens with the launch_rtscanner.sh script that might imply that repairnator will do this while running normally as well, which is less than desirable.

PR #785, opened 2018-11-05
fixed bug where createPR option was ignored in AbstractRepairStep

btellstrom:
Repairnator had been creating pull-requests even when specified not to (see #767 ). This PR aims to fix that. Also one change to the script "repair_buggy_build.sh" since the environment variables does not imply it will be used.


btellstrom:
It has been tested for cases `CREATE_PR=0` and `CREATE_PR=1` and works as intended.

btellstrom:
It has been tested for cases `CREATE_PR=0` and `CREATE_PR=1` and works as intended.

bloriot97:
Everything looks alright to me.

PR #786, opened 2018-11-28
Delete package-lock.json

surli:
This file should have not been pushed in the first place. We should add it in .gitignore. See: https://docs.npmjs.com/files/package-lock.json


surli:
NB: it should remove the security alerts we got.

monperrus:
Thanks a lot.

Issue #640, opened 2018-06-26
experiment: run NpeFix in fully automated manner for X patches (autonomous mode)


monperrus:
idea: run in fully automated manner for X patches, only with NpeFix (x=1, 5 10 in the beginning)

prerequisites:

* [x] https://github.com/Spirals-Team/npefix/issues/10
* [x] add support for automated pull request #30 
* [ ] write template of pull request title and body; see #704

monperrus:
in the autonomous mode, we need an option "stop after N submitted pull-request".

in the first trials, we will have N=1

monperrus:
Ongoing PR is https://github.com/Spirals-Team/npefix/pull/15

monperrus:
We need to change npeifx and then small stuff in maven-repair.

monperrus:
Per our conversation today

Four main milestones:

* send email
* create pull request
* integration with the scanner
* automatic stop after first pull-request

Example:

``` 
# exits after the first pull-request
# the rest of the configuration is either in shell script or as environment variable
./repairnator-autonomous.sh --max-pull-requests 1 

# for testing, we can focus on a single repo
./repairnator-autonomous.sh --max-pull-requests 1 --repo https://github.com/btellstrom/failingProject

# next with a repo list
./repairnator-autonomous.sh --max-pull-requests 1 --repo-list repo-list.txt

# finally with the RTScanner
./repairnator-autonomous.sh --max-pull-requests 1 --repo-mode RtScanner

```

PR #797, opened 2018-12-10
added maven-repair repository

btellstrom:
After an email-chain it was decided that we will, at least temporarily, move the [maven-repair](https://github.com/Spirals-Team/maven-repair/) into the Repairnator repository.


monperrus:
we have to remove .gitignore and .travis* 

and modify the existing travis config to also compile maven-repair.

bloriot97:
Oh it is a big commit, is there anything I need to particularly have a look on?

btellstrom:
@bloriot97 I think the most important thing is that travis-ci is properly set up, with a specific process for repairnator, and a separate process for maven-repair. So, if you check the build in travis you will see that there are two separate processes, from two different scripts. The relevant bit is that these scripts are not odd and that the `matrix` part of `.travis.yml` is properly set up. Since the build is passing after running the testsuite for about as long as usual I'd say it is most likely complete, but a second opinion is also nice.

[Maven-repair](https://github.com/Spirals-Team/maven-repair/) is another project, which for the moment has been transferred to this repository instead (partly for proceeding with #640 ). Also, most of the tools in that repository (if not all) are used in repairnator.

bloriot97:
Ok, it seems good for me. 
Thank you ;)

PR #806, opened 2019-01-30
Issue#804

### LINKED ISSUE #804: Prefix automatically generated emails by [repairnator]
Today the emails are prefixed by the name of the tool that generated the fix, e.g., [NopolAllTests]
It would make sense to prefix all emails by [repairnator] and mention the name of the tool later in the email object or email body
what do you think?

btellstrom:
In #804 it was suggested that the email reporting patches should be prefixed by repairnator instead of the toolname, and instead mention the toolname later in the email. This PR seeks to accomplish that.

Closes #804.


btellstrom:
Checked out the wrong branch.

PR #809, opened 2019-02-08
Deploy travis in stages to enable auto-deployment of docker-image

btellstrom:
By default travis runs the `after_build` after every execution of a script. This breaks the build since we can not deploy an image before we have built the Repairnator repo, which is larger and and thus takes longer time to complete.


btellstrom:
This configuration seems to do what is intended.

PR #817, opened 2019-03-06
Fix summary emails

### LINKED ISSUE #810: Bug: Summary emails are not being sent
With #799 repairnator gained the ability to send summary emails, however this does not seem to happen the way it should. This should be investigated and fixed.

btellstrom:
Preliminary changes to fix #810

Made some change that intend to fix the issues with summary emails. Some of it is currently wokring.

However, we still need to:

* [x] Make the regex for mongodb query work (Replaced with better solution based on `status`-field)

* [x] Make the list of repairtools be properly printed

Example of current email:
```
Summary email from Repairnator.

This summary contains the operations of Repairnator between Sat Jan 13 23:20:59 CET 2018 and Wed Mar 06 15:21:59 CET 2019.
Since the last summary Repairnator has:

Number of analyzed builds: 365663
Number of repair attempts made: 0
Total number of patches: 16919
Total number of patches found by NopolAllTests,AstorJMut,NPEFix: 0
```


monperrus:
remove sentence "since ..."
 
Total number of patches -> Total number of builds repaired with at least one patch.

btellstrom:
Got the list of tools to be properly read, was a slight mistake made previously when loading the configuration which was automatically remedied in everything else and thus had no effect up until now.

PR #820, opened 2019-03-15
High-level overview of Repairnator with basic list of references in the doc folder

fredloiret:
Just a basic overview of Repairnator's references that we provide to non technical people to whom we pitch the idea behind repairnator for the first time.


monperrus:
Thanks Frédéric!

PR #824, opened 2019-03-19
Fixes wrong character for formating years in strings

btellstrom:
Some documents, especially does close to the year shift, had the wrong datestrings, although the correct actual date. This caused weird behaviour in the graph on the dashboard, and this is not how it is supposed to act. 

This PR fixes the format in java that seems to have caused this behaviour.


monperrus:
Thanks!

PR #830, opened 2019-04-10
Fix output patch on console for NPEFix

### LINKED ISSUE #827: maven-repair:npefix should output the patch on the console
[maven-repair](https://github.com/Spirals-Team/repairnator/tree/master/maven-repair) has different modes.

Some modes output the patch to the console (eg `mvn repair:nopol`).

`mvn repair:npefix` does not output the patch to the console, and instead writes to a json file in folder `target`.

maven-repair:npefix should output the patch on the console

shrin18:
Gave internal call to the Json Object and got output on the Console.

Fix #827

### DIFF
--- a/maven-repair/pom.xml
+++ b/maven-repair/pom.xml
@@ -101,7 +101,7 @@
         <dependency>
             <groupId>fr.inria.spirals</groupId>
             <artifactId>npefix</artifactId>
-            <version>0.8-SNAPSHOT</version>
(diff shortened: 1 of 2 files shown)


shrin18:
I do not know why the console output is different from Intellij. I can show you the output on my laptop when I come to Lab today. I am currently in college for a course project

monperrus:
Thanks a lot for this first contribution

PR #837, opened 2019-04-15
fix: provide sensible default values for RtScanner

monperrus:
so as to run RtScanner without having to read the documentation

### DIFF
--- a/repairnator/repairnator-core/src/main/java/fr/inria/spirals/repairnator/LauncherUtils.java
+++ b/repairnator/repairnator-core/src/main/java/fr/inria/spirals/repairnator/LauncherUtils.java
@@ -326,6 +326,7 @@ public class LauncherUtils {
         opt.setShortFlag('n');
         opt.setLongFlag("name");
         opt.setStringParser(JSAP.STRING_PARSER);
+        opt.setDefault("spirals/repairnator:latest");
         opt.setHelp("Specify the docker image name to use.");
         return opt;
     }
--- a/repairnator/repairnator-core/src/main/java/fr/inria/spirals/repairnator/LauncherUtils.java
+++ b/repairnator/repairnator-core/src/main/java/fr/inria/spirals/repairnator/LauncherUtils.java
@@ -4,7 +4,6 @@ import com.martiansoftware.jsap.FlaggedOption;
 import com.martiansoftware.jsap.JSAP;
 import com.martiansoftware.jsap.JSAPResult;
 import com.martiansoftware.jsap.Switch;
-import com.martiansoftware.jsap.stringparsers.FileStringParser;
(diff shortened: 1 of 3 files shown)


monperrus:
@btellstrom the CI failure seems to be unrelated to this change: https://travis-ci.org/Spirals-Team/repairnator/jobs/521297233

Since it's related to dates, it may be a flaky test. WDYT?

monperrus:
ready for review. the goal is to be able to run the Scanner only as fast as possible (I spent hours debugging the parameters one after the other, finding out that they were not necessary in a default run).

repairnator/repairnator-core/src/main/java/fr/inria/spirals/repairnator/LauncherUtils.java
```diff
@@ -332,7 +326,6 @@ public static FlaggedOption defineArgDockerImageName() {
         opt.setShortFlag('n');
         opt.setLongFlag("name");
         opt.setStringParser(JSAP.STRING_PARSER);
-        opt.setRequired(true);
```
btellstrom: Since the `required` flag is removed we should probably set a default of `spirals/repairnator:latest`.

repairnator/repairnator-core/src/main/java/fr/inria/spirals/repairnator/LauncherUtils.java
```diff
@@ -113,30 +112,21 @@ public static FlaggedOption defineArgInput(String helpMessage) {
...
-        FileStringParser fileStringParser = FileStringParser.getParser();
-        if (launcherType == LauncherType.SCANNER || launcherType == LauncherType.CHECKBRANCHES) {
-            fileStringParser.setMustBeFile(true);
-        } else {
-            fileStringParser.setMustBeDirectory(true).setMustExist(true);
-        }
-        opt.setStringParser(fileStringParser);
-
-        if (launcherType == LauncherType.DOCKERPOOL || launcherType == LauncherType.REALTIME || launcherType == LauncherType.CHECKBRANCHES) {
+        opt.setDefault("/tmp/repairnator-output");
```
btellstrom: I am not sure exactly what this change would do, today when we run the RTScanner logs are created in the `$HOME` directory of the user running the RTScanner, and output is not to a single file, but rather there is log files for each container run as well as for the RTScanner itself. As I understand these changes, they would not allow this behaviour?

btellstrom: Realised this might be mitigated by the changes below, but I will keep this comment for now.

btellstrom:
Just realised that the `jars.md` file contains information that is quite vital to properly run Repairnator, so it might be worth renaming it to make it a bit more clear that that specific file is perhaps the first file one should look at?

PR #838, opened 2019-04-17
Number of patches option

### LINKED ISSUE #703: feat: add an argument to stop repairnator after X patches created
In order to complete #640 we need an argument to stop repairnator when X patches have been generated.
It means that we need to communicate between the RTScanner and the docker containers to get the info when a patch has been created.

btellstrom:
For the autonomous trial we want Repairnator to be able to stop after a set number of builds have been patched (or we have created a set number of patches, the functionality is there for that as well). This PR aims to create such a feature.

This is a partial step in #640.

Also closes #703


monperrus:
Diff seems too big. Rebase with master?

See also CI failure on Travis :-)

monperrus:
Thanks.

Issue #843, opened 2019-05-21
add support to gather all patches produced by different tools in a common list


monperrus:
we need to put all the patches gathered by different tools in a common list, so that we can do a common program analysis on all of them, in particular overfitting ranking.

this is closely related to #842

ycaxgjd:
Done. https://github.com/Spirals-Team/repairnator/pull/849

SophieHYe:
Hi @monperrus @ycaxgjd @btellstrom 
After review, this task is a bit ambiguous.
If the issue only focuses on providing a list of generated patches for the respirator pipeline, the task had been completed before and Jian added a supplemented JUnit test for it. 

Some comment from me:
I was expecting a bit more about this issue which is similar to this [script](https://github.com/Spirals-Team/repairnator/blob/134d7db9ff85bf9934b5f26c2a9fc6b2cf3705b2/repairnator/scripts/patched_builds.py ), it provides the functions: 
(1) query build IDs in a specific date range.
(2) for each build ID, query the generated patches for it. 
Maybe we can base on this script to format the return patches as a list and group by tools.  As @btellstrom suggested, having it in java and auto-updating it would certainly be a benefit.

monperrus:
per our discussion today, everything is OK

PR #849, opened 2019-06-11
add support to gather all patches produced by different tools in one …

ycaxgjd:
…common list

### DIFF
--- a/repairnator/repairnator-pipeline/src/test/java/fr/inria/spirals/repairnator/process/inspectors/TestJobStatus.java
+++ b/repairnator/repairnator-pipeline/src/test/java/fr/inria/spirals/repairnator/process/inspectors/TestJobStatus.java
@@ -1,14 +1,13 @@
-package fr.inria.spirals.repairnator.process.step.repair;
+package fr.inria.spirals.repairnator.process.inspectors;
 
 import ch.qos.logback.classic.Level;
 import fr.inria.jtravis.entities.Build;
 import fr.inria.spirals.repairnator.BuildToBeInspected;
+import fr.inria.spirals.repairnator.process.step.repair.NPERepair;
 import fr.inria.spirals.repairnator.process.step.repair.nopol.NopolSingleTestRepair;
 import fr.inria.spirals.repairnator.utils.Utils;
 import fr.inria.spirals.repairnator.config.RepairnatorConfig;
 import fr.inria.spirals.repairnator.process.files.FileHelper;
-import fr.inria.spirals.repairnator.process.inspectors.ProjectInspector;
-import fr.inria.spirals.repairnator.process.inspectors.RepairPatch;
 import fr.inria.spirals.repairnator.process.step.StepStatus;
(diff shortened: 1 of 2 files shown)


ycaxgjd:
Hey, this is for [issue #843](https://github.com/Spirals-Team/repairnator/issues/843)

btellstrom:
This only seems to be a test-file, is there something missing?

Also, there is currently a script that essentially does the [same thing](https://github.com/Spirals-Team/repairnator/blob/134d7db9ff85bf9934b5f26c2a9fc6b2cf3705b2/repairnator/scripts/patched_builds.py) (I missed the issue), but having it in java and auto updating it would certainly be a benefit. If you commit the function itself I'd be more than willing to have a look at it!

btellstrom:
Also, the way our test suite is working at the moment your build should always fail, so it is most likely not due to your addition, but rather because you are working from a fork.

ycaxgjd:
The gathering function should have already been implemented, so I only commit one corresponding test file.

Here are related code:
https://github.com/Spirals-Team/repairnator/blob/master/repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/step/repair/AbstractRepairStep.java#L95

https://github.com/Spirals-Team/repairnator/blob/master/repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/inspectors/JobStatus.java#L264

https://github.com/Spirals-Team/repairnator/blob/master/repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/inspectors/JobStatus.java#L268

btellstrom:
OH! It was per build basis, my mistake! Now I understand.

You have a lot of commented out code in the file. Is that supposed to be there? Also, not sure what the general practice is, but should this be in its own file? It seems that a feature like this should be tested in the same file as the class that creates the patchlist. WDYT?

ycaxgjd:
Hey, Benjamin, you are right. I made some changes, making it more reasonable.

btellstrom:
This looks good, now to the issue of making sure that the CI passes as well... 

Also, are you sure that NPEFix only has 4 patches? I usually get about 1270 when running it on the failingProject.

ycaxgjd:
Yes, 4 patches. I can pass this test file in my development environment. I guess that one kind of Nopol repairs usually gets over 1000 patches.

I am not sure the cause of this CI issue, maybe we could adjust the travis_wait parameter.

btellstrom:
> Yes, 4 patches. I can pass this test file in my development environment. I guess that one kind of Nopol > repairs usually gets over 1000 patches.
Okay, guess I remembered wrong.

Well, I know why the CI is failing most likely, it is because you are working from a fork which means that a lot of our tests can not run properly. I'm just curious how to resolve that issue. I think we'll speak about this later today (worst case scenario, I'll just create a branch that is a copy of your fork, run it through the CI and merge this PR if it passes).

btellstrom:
#853 Passes so this must also pass. Merged.

PR #850, opened 2019-06-13
add query scripts

btellstrom:
Changes the patched_builds script to print out all patches into a html format. Also adds some documentation and more interestingly some xpath specifications to get some interesting data.


monperrus:
thanks Benjamin.

PR #854, opened 2019-06-18
Add NPEFixSafe to repairnator 

btellstrom:
Currently, we have a mode for NPEFix that we call safe, but to properrly introduce it into Repairnator some more things need to be done.


monperrus:
Cool to have progress here!

monperrus:
Great step towards the autonomous mode!

Issue #860, opened 2019-07-11
Faling test RTScanner requires new repo


btellstrom:
The test `testRepositoryWithoutCheckstyleIsInteresting` is currently failing since the (repo)[https://github.com/nosan/embedded-cassandra] has recently added checkstyle. We should find a new repo which fulfills the same criteria as this one used to do. More specifically it should:

* Be on travis
* Be written in Java
* Not have checkstyle enabled

(The list above might not be absolute and I might have missed some details).

fermadeiral:
Hi, @btellstrom.

Repairnator will be always prone to have this type of failing test because the repo being used in the test can change. Finding another repo will only fix the test temporarily.

A suggestion that I have to give you is to checkout a commit from the repo, just after cloning it, where the checkstyle still didn't exist, so everything should be ok in the test.

Best,
Fernanda

monperrus:
One way is to fork https://github.com/nosan/embedded-cassandra under the organization repairnator and force master to be at the right commit. The fork won't evolve anymore.

btellstrom:
Hi @fermadeiral.

That is a very good point that I had not considered. However, I do think that Martins solution may be the easier here since the test is reliant on the slug to get the latest travis-build (might also be me not familiar with how to do it).

@monperrus I do not believe I am part of the repairnator organization, want to add me or should I just fork it myself?

monperrus:
you're now a member!

fermadeiral:
+1 to Martin's solution

PR #861, opened 2019-07-11
Add PullRequestSerialized

### LINKED ISSUE #859: Add pullrequest collection to the database
Since the aim is for Repairnator to create pull requests automcatically, it would be useful if we could easily get information about all pull requests it has created. Since we already have a database, the easiest way of doing this would be through a new collection in the database which we can quarry in much the same way as others.

btellstrom:
Adds a new class which serialized pull request so that we may easily extract data about the number of pull-requests Repairnator has created.

Closes #859


btellstrom:
Oh right, the CI will fail here due to #860 but there are no new tests (might want to add them, but did not find any existing ones of a similar sort).

repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/serializer/PullRequestSerializer.java
```diff
@@ -0,0 +1,74 @@
...
+import com.google.gson.JsonObject;
+
+import fr.inria.jtravis.entities.Build;
+import fr.inria.spirals.repairnator.process.inspectors.ProjectInspector;
+import fr.inria.spirals.repairnator.serializer.engines.SerializedData;
+import fr.inria.spirals.repairnator.serializer.engines.SerializerEngine;
+import fr.inria.spirals.repairnator.utils.DateUtils;
+
+/**
+ *
```
monperrus: add two sentences of explanation? when one uses this? for which goal?

monperrus:
ci still failing?

btellstrom:
Both this and #864 fail due to log length, but they should both have been merged with master, so I am really not sure what is the issue here...

PR #814, opened 2019-03-05
WIP: Deploy Repairnator in Kubernetes #813

gluckzhang:
Ongoing work on Issue #813

Issue #813, opened 2019-02-26
high availability: deploy Repairnator in Kubernetes


gluckzhang:
Core design:

- A queue system like ActiveMQ to save build ids
- A K8s cron job running repairnator scaner to get latest build ids and add them into the queue
- A set of pods running repairnator pipeline which fetches build id from the queue and does the analysis

Such an architecture could decrease coupling and take advantages of Kubernetes features. Since Repairnator has already implemented different Docker images, it is not very difficult to push forward.

Tasks (one per pull requests):

* [x] Rename `BuildRunner` into `DockerPipelineRunner`
* [x] Create empty `ActiveMQPipelineRunner extends AbstractPoolManager` (branch by abstraction strategy)
* [x] Change field name of `DockerPipelineRunner` to  `pipelineRunner`
* [x] Add interface PipelineRunner for ActiveMQPipelineRunner to implements
* [x] Implement `ActiveMQPipelineRunner`
* [x] Add `NopeRunner` which does nothing. It would allow to run the scanner only, even without docker installed.
* [x] `NopeRunner` should be the default
* [x] Add support for configuring the runner to use (ie either `DockerPipelineRunner` or `ActiveMQPipelineRunner`)
* [x] Change Travis CI config to deploy an ActiveMQ there and test the core functionality

btellstrom:
>* A queue system like ActiveMQ to save build ids
>* A K8s cron job running repairnator scaner to get latest build ids and add them into the queue

I believe these two things are already accomplished by the rt_scanner, which may be run indefinitely if one does not specify a duration.

Instead, I think that what needs to be done is to create a dockerimage with the rt-scanner. If I'm not completely wrong, we have the ability to run docker-containers remotely from rt-scanner? Which should allows us to create a set of pods to listen to the scanner. If I'm wrong about this, I know for a fact that we currently have an option to specify how many threads we wish to have running docker-containers, which we should be able to translate into instead communicating with Kubernetes.

gluckzhang:
Yes, we need to create a dockerimage for the rt-scanner :)
[...]

Issue #744, opened 2018-10-11
Travis is not working with regular pull requests


monperrus:
It's confusing to contribute to Repairnator, because CI fails for external pull-requests. 

Spin-off of @fermadeiral comment: https://github.com/Spirals-Team/repairnator/issues/579#issuecomment-396992772

fermadeiral:
First trial could be the long term view described in https://github.com/Spirals-Team/repairnator/issues/579#issuecomment-397203137

monperrus:
with the community expansion, this is becoming super high priority.

monperrus:
Fixed, #983 is a proof.

PR #862, opened 2019-07-11
[WIP: debugging]Add repairnator-pipeline dockerimage for k8s deployment (#813)

Tailp:
Hi, This is a PR breakdown in reference to this issue (#813) and original PR (#814) for deploying repairnator on Kubernetes. This PR only contains the pipeline image and some documentation :) .


repairnator/docker-images/kubernetes/README.md
```diff
@@ -0,0 +1,141 @@
+# Deploy a high avaliable Repairnator in Kubernetes
+
+*On going pull request*
+
+In order to make Repairnator highly available, it might be good to have configurations and tools to deploy Repairnator in Kubernetes, a preliminary idea here:
+
+- A queue system like ActiveMQ to save build ids
+- A K8s cron job running repairnator scaner to get latest build ids and add them into the queue
```
btellstrom: Why use the scanner instead of the realtime scanner? That would make the cron job unnecessary.

repairnator/docker-images/kubernetes/README.md
```diff
@@ -0,0 +1,141 @@
...
+To access to the web interface
+* kubectl get pods 
+It should look like activemq-XXXXXXX-XXXXX. Then 
+* kubectl exec -it activemq-XXXXXXX-XXXXX bash
+To expose to localhost for webinterface access and also publishing(sending message to queue) outside the cluster. 
+* kubectl port-forward activemq-XXXXXXX-XXXXX 8161:8161 61613:61613
+Access webinterface with "http://localhost:8161/admin/" in your browser. The default username is "admin" and password is also "admin".
+
+To send message to queue use the publisher.py script in queue-for-buildids folder. Syntax
+* python publisher.py -d queue-name message-1 message-2 message3 ... message-N
```
btellstrom: If I understand correctly you will make the scanner use this script to publish a message to the queue, but if this is the case, why not extend the scanner class to simply be able to communicate with the queue directly rather than taking this intermediate step?

repairnator/docker-images/kubernetes/pipeline-dockerimage/Dockerfile
```diff
@@ -0,0 +1,61 @@
+FROM maven:3.6.1-jdk-8
```
btellstrom: How is this docker-file different to the current pipeline image? (I see the stomp installation, but why would this be necessary?).

repairnator/docker-images/kubernetes/README.md
Tailp: That's one of the old information part . There are no cron job used in here :) so that part of the readme need to be removed . Then again the requirement for my task was minimal change or no change to the source code and after some checking on the RT scanner it seems that it uses docker images to validate/check the builds it found which is unfavorable since it would mean that we will be running docker inside docker. So was our thoughts about this after My own discussion with @Gluzhang

btellstrom: > Then again the requirement for my task was minimal change or no change to the source code and after some checking on the RT scanner it seems that it uses docker images to validate/check the builds it found which is unfavorable since it would mean that we will be running docker inside docker.

Wouldn't this be mitigated by running the queue system and starting up pods? Or am I misunderstanding how K8S work?

repairnator/docker-images/kubernetes/README.md
Tailp: Again this is only needed for deploying on kubernetes and in the normal case of repairnator there are no request of using activeMQ so according to the "minimal changes to the source" requirement it would not be wise for me to make changes in there. Also  if people want to manually test this out the need to manually inject buildid into the queue for testing.

repairnator/docker-images/kubernetes/pipeline-dockerimage/Dockerfile
Tailp: Stomp is for activeMQ without it there will be no queuing :) .

repairnator/docker-images/kubernetes/README.md
btellstrom: Okay, so you did not want to add a new class either? To me that would seem sensible, but I am not the one who gave you the task so not for me to say really.

repairnator/docker-images/kubernetes/README.md
Tailp: No, it seems we won't be able to escape docker container running inside the RTscanner pod container if we deploy it. The queue is only for putting jobs but the app itself will do the same thing inside the pod, which is running images. :)

btellstrom: Ahh, okay. Then I get it. Thanks for the response.

repairnator/docker-images/kubernetes/README.md
Tailp: Yea, I got a set of requirements and I am working in accordance to the supervision to avoid messing up. Then again it's logical since it would be wise that the deploying part should be as decoupled as possible to the source so if there are bugs on only my scripts or repairnator's code is needed for checking and we don't need to worry about the other might mess this up.

repairnator/docker-images/kubernetes/pipeline-dockerimage/Dockerfile
btellstrom: So you keep a queue inside of the pipeline? Wouldn't that potentially mess with state in the container?

repairnator/docker-images/kubernetes/README.md
Tailp: No problem man :).

btellstrom:
Since you are not adding any new java tests we should be comfortable ignoring the failing CI (and spare me the trouble of copying your fork to check the CI).

repairnator/docker-images/kubernetes/README.md
btellstrom: I would claim that a new class would accomplish the same regarding bugs at least, but okay.

repairnator/docker-images/kubernetes/pipeline-dockerimage/Dockerfile
Tailp: Ah no, the stomp is required to communicate with the global queue on the queue server. The queue is not in the pipeline but on activeMQ server.

btellstrom: Okay, so it is just to say "Hey, I'm done and ready for my next job"?

Tailp: Yes , basicly it is more desirable that way since we don't need to think about which job to assign to whom :).

monperrus:
Good progress! It's good that we start with a pull-request with this new folder `kubernetes`, and later in a subsequent PR, we'll change the way repair attempts are made (today calling docker is hardcoded in RtScanner which calls DockerPool).

Tailp:
Yes :) , I'll put the rest of the contents in our original [fork](https://github.com/gluckzhang/repairnator/tree/kubernetes) after this PR is merged .

monperrus:
I propose to rename folder `kubernetes` into `repairnator/kubernetes-support`

Tailp:
Hi @monperrus @btellstrom , are the tests for repairnator still flaky ? :) . I have done 6 attempts with different locations of my "kubernetes" folder. the different between the first "debug attempt" resulted to success is no different than the "debug attempt nr 6" and the last one failed despite being the same. It seems like the faulty is not due to my folder. :)

monperrus:
it should be solved, please rebase against master.

Tailp:
Hi the problem has to do with this issue (#744) explained by @btellstrom in this [PR](https://github.com/Spirals-Team/repairnator/pull/866). So it has nothing to do with flaky test or my own commits :) . Also, I'll make a new PR for this since this PR has become a bit dirty.

PR #864, opened 2019-07-12
Introduce PullRequestCounter

btellstrom:
We do not want repairnator to run for a set amount of `PATCHED` builds, rather we want it to create a set amount of pull requests. This PR changes the previous `PatchCounter` class to accomplish this.


monperrus:
- we keep field numberOfPatchedBuilds in RepairnatorConfig
- we keep the command line option change in RTLauncher
- we keep the PR based stopping behavior in RTScanner
- we keep PatchCounter and we add a new PullRequestCounter

monperrus:
ci still failing?

Tailp:
Hi :) , I just want to input that this fails because `BuildRunner` class name has been changed to `DockerPipelineRunner`. If you want to run RTScanner you need to first set the config with DOCKER pipelinemode in RepairnatorConfig :) , then everything will be proceeding as usual as prior the changes.

monperrus:
finally merged with some last changes to make CI pass. thanks.

PR #881, opened 2019-07-24
New conf for jGenProg

### LINKED ISSUE #880: question: does the pipeline stop after the first successful repair attempt?
For weeks, we mostly have kali patches from NopolAllTests.

This is boring and surprising because the other tools should also find some patches, starting with NpeFix and also with the kali patches of jGenProg.

Why do we have no jGenProg-kali patches?

Does the pipeline stop after the first successful repair attempt?

(anyway, we may disable NopolAllTests for now to debug the other tools)

monperrus:
fix #880

We'll start running multiple instances of Repairnator, one per repair tool (instead of having one single instance which runs all tools in a row)


Tailp:
No issue detected :) . Merged ?

monperrus:
We will push this in a private repo with all the passwords: https://github.com/kth-tcs/repairnator-configuration

PR #900, opened 2019-07-30
[WIP] reuse original-pipeline docker image

Tailp:
Hi :) , I implemented MessageListener interface and now ActiveMQPipelineRunner can be ran continuously as a message server. This is used to replace the python ActiveMQ to skip using python in this repo.


repairnator/repairnator-pipeline/pom.xml
```diff
@@ -90,6 +90,12 @@
             <version>1.5.0</version>
             <scope>test</scope>
         </dependency>
+        <!-- https://mvnrepository.com/artifact/javax.jms/javax.jms-api -->
```
monperrus: do we need that? it should be transitively gotten from rtscanner

Tailp: Hi @monperrus , it appears in the pom file we only import repairnator-core, but I'll try that out without importing :)

Tailp: I tried without importing and it does not work. We can do it this way or import these dependencies in repairnator-core pom.xml (but I don't think this is a good idea). Maybe you have a better idea how I should do this effectively :) ?. Thanks .

Issue #896, opened 2019-07-30
feat: add support for pushing into ActiveMQ in ProjectScanner


monperrus:
for @Tailp

Tailp:
@monperrus @gluckzhang hi :) , to dockerize this I need help deploying the new pipeline and project scanner component to our maven [repository](https://mvnrepository.com/artifact/fr.inria.repairnator), since I am not authorized.

monperrus:
> I need help deploying the new pipeline and project scanner component to our maven repository
> <https://mvnrepository.com/artifact/fr.inria.repairnator>
this is done automatically when merged to master (job deploy in Travis).

Tailp:
> I need help deploying the new pipeline and project scanner component to our maven repository <https://mvnrepository.com/artifact/fr.inria.repairnator>
> this is done automatically when merged to master (job deploy in Travis).

Hi :) , but from the look of it the most recent one is from last year Aug, 2018 for the scanner. <https://mvnrepository.com/artifact/fr.inria.repairnator/repairnator-scanner> . also when I test downloading the latest version into a jar and run it, it does not contain the new options we merged yesterday.

monperrus:
The SNAPSHOT is the latest: https://oss.sonatype.org/content/repositories/snapshots/fr/inria/repairanator/repairnator-pipeline/

PR #901, opened 2019-07-31
Feat: add empty ScannerBuildListener, options and dependencies

Tailp:
Hi :), related issue ( #896). PR contains new options for project scanner and dependencies import in pom and an empty class to be implemented.


monperrus:
The small PR is good, thanks. But the three changes are unrelated to each other.

If you want to add a listening feature to the Scanner, you would first add one interface, and at least one call to the interface method somewhere.

Tailp:
Hi :) , it will be implementing an interface provided by ActiveMQ later called MessageListener. So I think we don't have to write our own for that :) .

Tailp:
#904 need to be closed first before this continues :) .

Tailp:
@monperrus close to open a new PR ? :) .

repairnator/repairnator-scanner/src/main/java/fr/inria/spirals/repairnator/scanner/ScannerBuildListener.java
```diff
@@ -0,0 +1,3 @@
+package fr.inria.spirals.repairnator.scanner;
+
+public class ScannerBuildListener {}
```
monperrus: this is out of the blue: it should be used somewhere. make it an interface?

Tailp:
Resolved :) .

repairnator/repairnator-core/src/main/java/fr/inria/spirals/repairnator/BuildListener.java
```diff
@@ -0,0 +1,7 @@
+package fr.inria.spirals.repairnator.pipeline;
+
+import javax.jms.MessageListener;
+/**
+ * This class fetch build ids from ActiveMQ queue and run the pipeline with it.
+ */
+public interface BuildListener extends MessageListener {}
```
monperrus: this interface is never used

Tailp:
@gluckzhang @monperrus , thank you :) . Awaiting for new comments.

gluckzhang:
Just went through relevant classes and found that:  
The procedure of scanning projects should be the same, the difference is:  
- print the list of builds to a file
- print the list of builds to stdout
- print (send) the list of builds to the queue

If this makes sense, I think we could just enhance the method `private void processOutput(Map<ScannedBuildStatus, List<BuildToBeInspected>> listOfBuilds)` as follows:

```diff
private void processOutput(Map<ScannedBuildStatus, List<BuildToBeInspected>> listOfBuilds) {
    if (this.config.getOutputPath() != null && !this.config.getOutputPath().isEmpty()) {
        this.printToFile(listOfBuilds);
+   } else if (this.config.getQueueURL() != null && !this.config.getQueueURL().isEmpty()) {
+       this.sendToQueue(listOfBuilds);
    } else {
        this.printToStdout(listOfBuilds);
    }
}
```

Then we implement another method which sends build ids to the queue. WDYT? @monperrus

monperrus:
> I think we could just enhance the method private void processOutput(Map<ScannedBuildStatus, List<BuildToBeInspected>> listOfBuilds) as follows:

Good idea.

Tailp:
Hi :) , sorry I failed to understand what to fix. Should I add the implementation now ? .

gluckzhang:
> Hi :) , sorry I failed to understand what to fix. Should I add the implementation now ? .

Hi, we could implement a method which sends build ids to the queue in class `Launcher` and update the `processOutput` method. See my previous comment, in this way you don't need to design new interfaces and you directly follow the original workflow :)

Tailp:
> 
> 
> > Hi :) , sorry I failed to understand what to fix. Should I add the implementation now ? .
> 
> Hi, we could implement a method which sends build ids to the queue in class `Launcher` and update the `processOutput` method. See my previous comment, in this way you don't need to design new interfaces and you directly follow the original workflow :)

Hi :) @gluckzhang, I have already implemented another way locally which more or less is similar to your suggestion and it also works after testing it out. Should I implement it in a new PR and let you check , if there are some problems then we can compromise or go with your suggestion, then perhaps we also will have a better view of other hidden bugs ? . WDYT ? :).

gluckzhang:
> Should I implement it in a new PR and let you check

No, the goal is to get this PR merged, though the design may need to be optimized :)

For the current design, I have some doubts:

1) The code does not show how you monitor the luncher in `ScannerBuildListener`. Probably in `runAsConsumerServer()`? But do you need to change the `Luncher` a lot?

2) If you would like to design an interface called `BuildListener`, then `ActiveMQListener` should be one of its implementation. I am not sure if `javax.jms.MessageListener` is specifically for ActiveMQ or it is a general one. If later we have another queue system supported by Redis, is the current design easily extended?

About polymorphism, [this PR (add support for different runner with polymorphism)](https://github.com/Spirals-Team/repairnator/pull/891) could be a good example for you. Let's discuss more tomorrow offline :)

gluckzhang:
@Tailp I just found that we are discussing two requirements :)

- Fetch project slug from the queue (before the inspection logic in `Launcher`)
- Send build ids to the queue (after the main logic in `Launcher`)

So let's combine our ideas and get this solved tomorrow.

Tailp:
@gluckzhang , actually not much changes :) . Only a new function called kubernetesProcess() will be used instead of the so called mainProcess() . KuberntesProcess(String slug) will reuse already-existed scanner functions and no unncessary steps like serializations and check a list. I will be a bit hard to read if we try to embed our codes in the current mainProcess, which I found out some weeks ago when I coded it :) .

Tailp:
> 
> 
> @Tailp I just found that we are discussing two requirements :)
> 
>     * Fetch project slug from the queue (before the inspection logic in `Launcher`)
> 
>     * Send build ids to the queue (after the main logic in `Launcher`)
> 
> 
> So let's combine our ideas and get this solved tomorrow.

Let's do that :) . I can just show you my implementation tomorrow :) . Yes the scanner works as both a server listener and a publisher which make it a bit more tricky.

gluckzhang:
> Yes the scanner works as both a server listener and a publisher which make it a bit more tricky.

Sounds a bit weird, the relevant issue (#896) is about add support to push build ids into the queue. Not sure if you need to keep the scanner runs continuously like `RTScanner`.

Tailp:
> > Yes the scanner works as both a server listener and a publisher which make it a bit more tricky.
> 
> Sounds a bit weird, the relevant issue (#896) is about adding support to push build ids into the queue. Not sure if you need to keep the scanner runs continuously like `RTScanner`.

Our implementation in python works that way :) . It listens for a slug in a scanner queue and then submit build ids after scanning to pipeline queue.

Tailp:
Per discussion today with @gluckzhang , we should implements MessageListener as a separate interface than extending BuildListener with it, in case if we want to implement BuildListener but not using ActiveMQ (MessageListener) :)

gluckzhang:
> Per discussion today with @gluckzhang , we should implements MessageListener as a separate interface than extending BuildListener with it, in case if we want to implement BuildListener but not using ActiveMQ (MessageListener) :)

Not exactly. I mean as an interface `BuildListener` designed by you, it should be loosely coupled, i.e. it should not depend on ActiveMQ or relevant interfaces. You could have one implementation: `ScannerBuildListener` (or maybe call it `ActiveMQBuildListener`), which communicates with a queue system and needs to implement `MessageListener` interface. If later you would like to have another implementation, for example, a `FileBuildListener` which reads information from a file, you only need to implement your own `BuildListener` interface.

repairnator/repairnator-core/src/main/java/fr/inria/spirals/repairnator/BuildListener.java
```diff
@@ -0,0 +1,6 @@
+package fr.inria.spirals.repairnator;
+
+/**
+ * This class fetch build ids from ActiveMQ queue and run the pipeline with it.
+ */
+public interface BuildListener {}
```
gluckzhang: Your interface should not be empty. According to your `ScannerBuildListener`, method `setLauncher` and `runAsConsumerServer` should be put here.

repairnator/repairnator-core/src/main/java/fr/inria/spirals/repairnator/BuildListener.java
```diff
@@ -0,0 +1,6 @@
+package fr.inria.spirals.repairnator;
+
+/**
+ * This class fetch build ids from ActiveMQ queue and run the pipeline with it.
```
gluckzhang: The interface is an abstract presentation, it should not be relevant to ActiveMQ. This line of comment refers to your `ScannerBuildListener`.

repairnator/repairnator-scanner/src/main/java/fr/inria/spirals/repairnator/scanner/ScannerBuildListener.java
```diff
@@ -0,0 +1,32 @@
...
+        this.launcher = launcher;
+    }
+
+    public static ScannerBuildListener getInstance() {
+        if (scannerBuildListener == null) {
+            scannerBuildListener = new ScannerBuildListener();
+        }
+        return scannerBuildListener;
+    }
+    public void runAsConsumerServer() {}
```
gluckzhang: If every implementation needs this method and the `setLauncher`, they should be put into the `BuildListener` interface.

repairnator/repairnator-scanner/src/main/java/fr/inria/spirals/repairnator/scanner/Launcher.java
```diff
@@ -346,6 +376,7 @@ private void printToStdout(Map<ScannedBuildStatus, List<BuildToBeInspected>> lis
 
     public static void main(String[] args) throws Exception {
         Launcher launcher = new Launcher(args);
+        ScannerBuildListener.getInstance().setLauncher(launcher);
```
gluckzhang: Usually we use the interface to declare an instance, the current line does not take advantages of polymorphism. For example:

```Java
BuildListener buildListener = null;
if (...) { // based on your arguments
    buildListener = new ActiveMQBuildListener();
} else if (...) {
    buildListener = new FileBuildListener();
}
if (buildListener != null) {
    buildListener.setLauncher(launcher);
    buildListener.runAsConsumerServer();
}
launcher.mainProcess();
```

Tailp:
@gluckzhang :) hi, please have a look at the latest commit.

Tailp:
@gluckzhang thanks for the comments :) . Resolved.

gluckzhang:
> @gluckzhang thanks for the comments :) . Resolved.

Thanks, LGTM!

monperrus:
Thanks a lot!

PR #914, opened 2019-08-02
refactor: pipeline mode is purely automated

monperrus:
@Tailp look at how we can even remove more if conditions


Tailp:
Nice :), I have I lot to learn. Thank you very much for showing me this :) .

PR #915, opened 2019-08-02
feat: a faster scanner

monperrus:
This introduces a super fast scanner directly inspired from @tdurieux 's travis-listener.

It produces 50x failing builds.

This means we need a way to scale the spawning of docker containers.


tdurieux:
Nice!

tdurieux:
It is also possible to do the same thing at the build level, the advantages are:
- direct access to know if it is a PR
- faster: less build compared to jobs

monperrus:
I don't understand; could you elaborate?

tdurieux:
You are scanning at the job level you could also scan at the build level.

monperrus:
Yes, interestingly, the builds endpoint was also closed in v2 in anonymous mode `curl --verbose --header "Accept: application/json; version=2" https://api.travis-ci.org/builds`

It means we first have to go through `https://api.travis-ci.org/jobs` anyway, then to get the most recent build and play with build numbers.

Do you agree?

tdurieux:
I dont really know what curl is doing but the v2 is not closed yet :)

you can try by yourself (for this url: https://api.travis-ci.org/builds?ids[]=108479361&ids[]=108479365):
```
curl --globoff --header "Accept: application/vnd.travis-ci.2.1+json, */*; q=0.01" https://api.travis-ci.org/builds\?ids\[\]\=108479361\&ids\[\]\=108479365
```

monperrus:
I confirm that `/builds` works in v2.

Interestingly, one can pass wrong ids as input, it would only return valid builds (so fewer elements answered).

tdurieux:
> Interestingly, one can pass wrong ids as input, it would only return valid builds (so fewer elements answered

It is the same behavior with `https://api.travis-ci.org/jobs`.

PR #941, opened 2019-09-02
Fix: try fixing travis error

Tailp:


Tailp:
it appears that the thing was just a bit flaky.

PR #964, opened 2019-10-15
Scan builds' code for SequenceR training data

javierron:
## Description

The `SequencerLearnerScanner` class is based on `FastScanner`, but re purposed to fetch passing builds and process them using `SequencerCollector`.

SequencerCollector will be responsible for fetching and scanning all passing builds' commits for compatible [SequenceR](https://github.com/KTH/sequencer) training data.

## Current Status

Passing java builds' commits are being retrieved from GitHub, the current filter checks for:
- Single-file commits
- File to actually be a .java file

## Next Steps
- Filter out test files
- Implement filter to only allow single-line fixes.
  - filter out patches with several points of modification (scan for multiple `@@` lines)
  - scan for pattern `-[line] \n +[line]` and validate that it is not a block change boundary. Use regex(?)
- Deal with data storage and tokenization(?)

## Problems
- Currently the TravisCI API is returning build information with inconsistent date formats as described [here](https://travis-ci.community/t/api-v3-build-buildid/5479). A work around has been made, but if this is intended behavior, or it is not fixed, it will be necessary to update the JTravis lib to deal directly with this inconsistency. 
UPDATE: TravisCI has responded saying that it _looks like_ a bug and that they will check on it _soon_, so I believe we can use the workaround until they give a more concrete response.

### DIFF
--- a/repairnator/repairnator-realtime/src/main/java/fr/inria/spirals/repairnator/realtime/BuildHelperV2.java
+++ b/repairnator/repairnator-realtime/src/main/java/fr/inria/spirals/repairnator/realtime/BuildHelperV2.java
@@ -1,0 +1,31 @@
+package fr.inria.spirals.repairnator.realtime;
+
+import java.util.Arrays;
+import java.util.Optional;
+import java.util.Properties;
+
+import fr.inria.jtravis.JTravis;
+import fr.inria.jtravis.TravisConstants;
+import fr.inria.jtravis.entities.v2.BuildV2;
+import fr.inria.jtravis.helpers.BuildHelper;
+
+
+// The motivation of this class is that currently the Travis API is returning inconsistent
+// date formats when requesting a build. Therefore this helper and corresponding entity
+// excludes all attributes that are not needed for the current purpose.
+public class BuildHelperV2 extends BuildHelper {
+
+	public BuildHelperV2(JTravis jtravis) {
+		super(jtravis);
+	}
+	
+	
+	public Optional<BuildV2> fromIdV2(long id) {
+		 Properties properties = new Properties();
+	        properties.put("include","job.config");
+
+	        return super.getEntityFromUri(BuildV2.class, Arrays.asList(TravisConstants.BUILD_ENDPOINT, String.valueOf(id)), properties);
+	    
+	}
+	
+}
--- a/repairnator/repairnator-realtime/src/main/java/fr/inria/spirals/repairnator/realtime/BuildHelperV2.java
+++ b/repairnator/repairnator-realtime/src/main/java/fr/inria/spirals/repairnator/realtime/BuildHelperV2.java
@@ -10,9 +10,10 @@ import fr.inria.jtravis.entities.v2.BuildV2;
 import fr.inria.jtravis.helpers.BuildHelper;
 
 
-// The motivation of this class is that currently the Travis API is returning inconsistent
-// date formats when requesting a build. Therefore this helper and corresponding entity
-// excludes all attributes that are not needed for the current purpose.
+/** The motivation of this class is that currently the Travis API is returning inconsistent
+ * date formats when requesting a build. Therefore this helper and corresponding entity
+ * excludes all attributes that are not needed for the current purpose.
+ */
 public class BuildHelperV2 extends BuildHelper {
 
 	public BuildHelperV2(JTravis jtravis) {
--- a/repairnator/repairnator-realtime/src/main/java/fr/inria/spirals/repairnator/realtime/FastScanner.java
+++ b/repairnator/repairnator-realtime/src/main/java/fr/inria/spirals/repairnator/realtime/FastScanner.java
@@ -46,7 +46,6 @@ public class FastScanner implements Runnable {
         int nInteresting = 0;
         Set<Long> done = new HashSet<>();
         JobHelperv2 jobHelperv2 = new JobHelperv2(RepairnatorConfig.getInstance().getJTravis());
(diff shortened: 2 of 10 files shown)


repairnator/repairnator-realtime/src/main/java/fr/inria/spirals/repairnator/realtime/BuildHelperV2.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.Optional;
+import java.util.Properties;
+
+import fr.inria.jtravis.JTravis;
+import fr.inria.jtravis.TravisConstants;
+import fr.inria.jtravis.entities.v2.BuildV2;
+import fr.inria.jtravis.helpers.BuildHelper;
+
+
+// The motivation of this class is that currently the Travis API is returning inconsistent
```
monperrus: use Javadoc `/**    ... */`

repairnator/repairnator-realtime/src/main/java/fr/inria/spirals/repairnator/realtime/FastScanner.java
```diff
@@ -46,6 +46,7 @@ public void run() {
         int nInteresting = 0;
         Set<Long> done = new HashSet<>();
         JobHelperv2 jobHelperv2 = new JobHelperv2(RepairnatorConfig.getInstance().getJTravis());
+        SequencerCollector collector = new SequencerCollector();
```
monperrus: a little but of inheritance would help: this should be in a subclass of FastScanner `SequencerLearnerScanner`

repairnator/repairnator-realtime/src/main/java/fr/inria/spirals/repairnator/realtime/SequencerCollector.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.Vector;
+
+
+import fr.inria.jtravis.entities.v2.BuildV2;
+import fr.inria.jtravis.entities.v2.JobV2;
+import fr.inria.spirals.repairnator.config.RepairnatorConfig;
+
+import org.kohsuke.github.*;
+
+public class SequencerCollector {
```
monperrus: one sentence API doc

repairnator/repairnator-realtime/src/main/java/fr/inria/spirals/repairnator/realtime/FastScanner.java
javierron: WDT about a Scanner abstract class?

javierron:
Working on it!

Checks failed because I did not had an eclipse account, and had not submitted the ECA thing.

monperrus:
would be great!

monperrus:
Still WIP? What are the next steps?

javierron:
# Update

## Current Status
- Rebased branch to current upstream:master
- Developing a solution for actually filtering out multi hunk patches.

## Next steps
- Storage solution, for now this solution is: plain text files on disk.
  
## Problems
- The GitHub API endpoint that is currently being used does not actually return the patches in "unified fomat" (example [here](https://api.github.com/repos/javierron/repairnator/commits/d23782e9d18bef4baa7c75ad1ea4c3eebe20639d)), The lines that have the filename are missing (`+++ /path/to/file`). This makes parsers like [java-diff-utils](https://github.com/java-diff-utils/java-diff-utils) to not function. Solutions to be explored follow:
   - Solution 1: Get the whole commit .diff file from elsewhere. although it seems that having to parse the whole diff file of a possibly multi-file commit does not seem very efficient, since using the current endpoint we can filter out multi-file commits quite easily.
   - Solution 2: Prepend `+++ /file/name` to the patch, and use `java-diff-utils`, this was tried an partially works, except that `java-diff-utils` does not implement the full spec for unified diff, and does not yield correct results. Actual solution would be to try another parser (maybe [xwiki-commons-diff](https://github.com/xwiki/xwiki-commons/tree/master/xwiki-commons-core/xwiki-commons-diff/xwiki-commons-diff-api))
   - Solution 3: Make a simple patch parser based on `java-diff-utils`, their implementation does not seem too complex and can be modified/extended.

javierron:
## Update

### Current Status
- Rebased branch to current upstream:master
- Filter for single line fixes is done (I will leave this running overnight to see if we can catch a fix or two)

### Next steps
- Error handling
- Refactor

repairnator/repairnator-realtime/src/main/java/fr/inria/spirals/repairnator/realtime/FastScanner.java
javierron: Since the scanners do not share a common interface I think it is ok to just leave both implementing runnable and not creating an abstract class just because. I think your point was for me to not modify FastScanner directly. Am I correct?

monperrus:
> I think your point was for me to not modify FastScanner directly.


Yes, thanks for the progress!

javierron:
Overnight scanning yielded about ~30 results. I think that's a pretty good number.

monperrus:
do you mean 30 instances of a one-line change in a Java file?

monperrus:
Nice. To maximize the amount of data, we can take all one liner changes (even if they are in commits
with bigger chhanges on other files), and we may even consider taking one-line hunks in the same
file if the hunks are far enough from each other in the file. WDYT?

javierron:
That makes sense, with what I have right now, it's not that hard to extend, and I can make a config to enable/disable filters.

monperrus:
Thanks a lot!

PR #993, opened 2019-12-12
Update documentation (close #976)

### LINKED ISSUE #976: Advice about README
This is just an advice about the README: the example command in the section [Fixing a build](https://github.com/eclipse/repairnator/blob/master/doc/README.md#fixing-a-build) (in which it is explained how to run repairnator-pipeline using its Docker image) has not the parameter GITHUB_OAUTH. 

Since without this parameter Repairnator generates some authentication errors,  in my opinion, it could be useful to add the parameter to the example (as reported in the [Repairnator.cfg section](https://github.com/eclipse/repairnator/blob/master/doc/repairnator-config.md#github_oauth)).

Finally, it could be useful to add a section in which to put all the environment variables that can be configured via command line when repairnator/pipeline is run with Docker.

dginelli:
Updated the documentation based on #976 and reviewed also the other sections.


monperrus:
Thanks a lot. Documentation PR are essential!

PR #994, opened 2019-12-12
Added a new Step Status

dginelli:
When a repair tool doesn't find any patches, Repairnator now uses the status `PATCH_NOT_FOUND` instead of `SKIPPED`.

Signed-off-by: ginellidavide@gmail.com <ginellidavide@gmail.com>


monperrus:
Thanks, it will help a lot to understand the output!

Issue #469, opened 2018-03-20
Repairnator for Jenkins: run repairnator when PR are coming or builds fail


surli:
Configuration space:

* Trigger
  * BS: Build step for Repairnator
  * Post-build action of "Freestyle job"
    * PBA-GH For Github builds (incl. PR builder)
    * PBA-GIT For any Git repo
  * NJT: New full-fledged new job type "Repairnator job"   
* Feedback:
  * Feedback-Jenkins: Patch stored and shown on Jenkins directly
  * Feedback-email: Patch sent over email
  * Feedback-Github: Comment/Pull-request on Github
  * Feedback-Gitlab:  Comment/Pull-request on Gitlab

surli:
I'm closing this one: I won't have time to fix the issues I had with jenkins. And we manage to work using only Travis.

monperrus:
Many industry partners use Jenkins, it would be great to have a Jenkins plugin for Repairnator.

monperrus:
The Jenkins plugin adds a new kind of "Post-build action", called "Run Repairnator upon failing builds".

henry-lp:
@monperrus Hi Martin , I just recently start working on this per discussion . 

The plan:
* Create a dummy plugin with post build [done]
* Introduce `repairnator-pipeline` as maven dependency into the dummy plugin [WIP]
* Host the plugin on Jenkins central repository

This way if the plan is possible, we also ease the CI/CD for the plugin when newer version of pipeline itself is deployed, since it's only  a maven dependency specified in the pom.xml for the plugin. That's my current design .

henry-lp:
Now to my current problem .

This issue happens when I tried to introduce `repairnator-pipeline` [maven dependency[(https://search.maven.org/artifact/fr.inria.repairnator/repairnator-pipeline/3.1/jar) into one of my dummy travis project [Henrylp/travisplay](https://github.com/henry-lp/travisplay) from Sonatype Nexus . Pom file and error output log are down below .

It seems that it unable to install the `repairnator-root` dependency inside the `repairnator-pipeline` dependency according to the err log . 

It would be nice If I know a fix or a workaround for this . Otherwise I will modify my plan a bit to avoid this :) .

My pom.xml
```
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
[...]

PR #995, opened 2019-12-18
Feat: new entry point method for pipeline from Jenkins plugin

henry-lp:
Issue (#469)

Hi @monperrus :) , This is a introduce a new entry point for the pipeline to be used by Jenkins plugin . It skip JSAP since it causes some errors and we don't really need JSAP if we use the plugin. I need this released to Sonatype later so I can import it as a maven dependency for our plugin.


henry-lp:
Close to sign off on eclipse.  New identical PR ( #996)

Issue #867, opened 2019-07-15
Travis CI failed for a fresh master fork


Tailp:
Hi, I test making a PR with only a comment line changed after forking from master branch and Travis failed while it's supposed to be passing . Maybe it's an issue on my part ?. Please have a look :) . 

Related PR : (#866)

Output: (more details are in this [Travis raw output log](https://api.travis-ci.org/v3/job/558999668/log.txt)):
```
Tests run: 9, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 113.068 sec <<< FAILURE! - in fr.inria.spirals.repairnator.scanner.ProjectScannerTest
testGetBuildToBeInspectedWithPassingWithPreviousPassingFromPR(fr.inria.spirals.repairnator.scanner.ProjectScannerTest)  Time elapsed: 33.525 sec  <<< FAILURE!
[...]

PR #1003, opened 2020-01-15
fix: does not output GitHub Token in logs

dginelli:
When Repairnator is configured to push the results on an external GitHub repository, the "repairnator.maven.testproject.log" file associated with some builds (e.g., this one: 361311603) contains the GitHub token used by the user. This causes a security problem, GitHub revokes the OAuth token, and thus it is necessary to generate another one every time this situation happens.

The proposed fix replaces the user GitHub Oauth token with the string "CONFIDENTIAL_INFORMATION" before writing it in the .log files.

You can see an example of log file generated with the proposed fix [here](https://github.com/dginelli/test-token-security/blob/tverdokhlebd-mined-in-361311603-20180402-211043/repairnator.maven.testproject.log#L95). The original version of the same file is [this one](https://github.com/dginelli/test-token-security-2/blob/tverdokhlebd-mined-in-361311603-20180402-211043/repairnator.maven.testproject.log#L95).

Signed-off-by: ginellidavide@gmail.com ginellidavide@gmail.com

### DIFF
--- a/repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/maven/output/MavenOutputHandler.java
+++ b/repairnator/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/maven/output/MavenOutputHandler.java
@@ -8,6 +8,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.FileWriter;
 import java.io.IOException;
+import java.util.regex.Pattern;
 
 /**
  * Created by urli on 15/02/2017.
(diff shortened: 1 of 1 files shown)


monperrus:
Nice catch! Thanks a lot!

monperrus:
It seems that this breaks a test. Any idea?

henry-lp:
@monperrus maybe this ? #867

dginelli:
I changed the way used to find the GitHub token in .log files, because I found out another build (this one 353063469) that saves the token in [another way](https://github.com/dginelli/repairnator-experiments-jkali-test/blob/4c25122b7304bdbb0143f0cc879130d04aa193ac/repairnator.maven.testproject.log#L73), different from the previous cases.

monperrus:
thanks. we still have a failing test case to understand.

monperrus:
Thanks a lot Davide!

PR #1011, opened 2020-01-28
Fix project identifier folders

dginelli:
Changed the folder names associated with the builds created by Repairnator and Astor in their workspace to avoid conflicts. More details are in the commit messages.


monperrus:
Thanks a lot!

PR #1013, opened 2020-01-28
chore: remove checkstyle

monperrus:
because it's not enforced in CI and we now have "154 errors reported" at release/deployment time.


gustavopinto:
Great! I'd suggest the same thing.

PR #1016, opened 2020-01-29
feat: Sequencer collector now stores hunks separately

javierron:
Single-change hunks are now stored separately to easily tokenize them.

Also disabled commit/push to data repository. If we need to start storing data again it can be enabled. 


monperrus:
Thanks!

PR #1030, opened 2020-02-13
fix: pipeline ActiveMQ's access control args

gluckzhang:
Previously the pipeline only supports to connect ActiveMQ without username and password. This commit fixes the issue.

The corresponding test case has been enhanced. The added arguments have been documented in `doc/repairnator-kubernetes.md` and `repairnator-pipeline.yaml`.

Signed-off-by: Long Zhang <zhanglong3030@qq.com>


monperrus:
Thanks a lot!

PR #1048, opened 2020-03-05
fix: a path error when restarting pipeline pods

gluckzhang:
The k8s related commands should be executed at the root folder of the project. This PR fixes the error `cp: cannot stat '.travis/kubeconfig.skeleton': No such file or directory` in travis. (e.g., https://travis-ci.org/eclipse/repairnator/jobs/658817214#L484)

Signed-off-by: Long Zhang <zhanglong3030@qq.com>


monperrus:
thanks!

Issue #1041, opened 2020-02-25
Jenkins plugin: add config parameter "email"


monperrus:
When using the Jenkins plugin in a non-Github plugin, the patch only appears in the log.

We can add an advanced config parameter "email": if a patch is found and we are not in a Github context, the patch is sent by email to the developers.

@henry-lp WDYT?

henry-lp:
Yes , I'll add this new feature. Is this urgent ?  

If I can start in one or two week after my schedule loose up a bit, it would be perfect. :)

monperrus:
Great, thanks Henry! No urgency here :-)

PR #1052, opened 2020-03-16
[WIP]: Jenkins plugin email patches

henry-lp:
This address issue (#1041) 

@monperrus  this feature can be tried at [link](http://jenkins.kth-assert.net/job/repairnator%20plugin%20demo/) :)

### DIFF
--- a/src/repairnator-jenkins-plugin/pom.xml
+++ b/src/repairnator-jenkins-plugin/pom.xml
@@ -58,6 +58,11 @@
     </dependencyManagement>
     <dependencies>
         <dependency>
+          <groupId>commons-io</groupId>
+          <artifactId>commons-io</artifactId>
+          <version>2.5</version>
+         </dependency>
+        <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>structs</artifactId>
         </dependency>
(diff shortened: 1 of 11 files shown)


monperrus:
Great to see progress here. Thanks henry!

Jenkins has its own configuration to send emails, used for example to send emails upon build failures.

This means that we can save the user from putting all the difficult and potentially sensitive SMTP info by using this service instead.

WDYT?

henry-lp:
hi @monperrus ,

Perhaps you mean [this feature](https://www.360logica.com/blog/email-notification-in-jenkins/), I read around and apparently you still need to provide all of those credentials there. 

There are also some problems using Jenkins email notifications, we use tmp folder to generate of stuffs so the folder name will not be fixed each time therefore one of the possibility is to set an enviroment variable for the email notification plugin to parse it after repairnator is done. Another is that the user provide a folder for repairnator to use , but then the problem is that that folder need to have enough priviledges which is not very handy either. All of this comes with an obvious price that we now have to depends on another plugin.

By far I don't see this solution being easier or more handy than providing those information directly into Repairnator, you only need to set up a fake account for repairnator to use. WDYT :) ?

monperrus:
I mean the buil-in support (https://stackoverflow.com/a/11896118): in all serious Jenkins instances,
there is already a configured SMTP server.

henry-lp:
new PR (#1056)

PR #1056, opened 2020-03-22
feat: support for email notification in Jenkins plugin

henry-lp:
This address issue (#1041) 

@monperrus  this feature can be tried at [link](http://jenkins.kth-assert.net/job/repairnator%20plugin%20demo/) :) 

old PR from fork ( #1052)


henry-lp:
@monperrus  ready for review.

I have changed so that we reuse config information from the built-in Email Notifier :) . 

Try out at (http://jenkins.kth-assert.net/)

src/repairnator-jenkins-plugin/src/main/java/io/jenkins/plugins/main/RepairnatorJarDownloader.java
```diff
@@ -25,11 +26,15 @@ public String getLatestJarUrl(String snapshotUrl) throws IOException{
...
+        System.out.println(doc);
+        System.out.println("LatestJarUrl: " + jarUrl  + latestJar);
+        return jarUrl +  latestJar;
     }
 
     public void downloadJar(String snapshotUrl) throws IOException{
-        String latestJarUrl = this.getLatestJarUrl(snapshotUrl);
+        /*String latestJarUrl = this.getLatestJarUrl(snapshotUrl);*/
+
+        String latestJarUrl = "https://repo.jenkins-ci.org/snapshots/fr/inria/repairnator/repairnator-pipeline/3.3-SNAPSHOT/repairnator-pipeline-3.3-20200316.204020-11-jar-with-dependencies.jar";
```
monperrus: hard-coded?

src/repairnator-jenkins-plugin/src/main/java/io/jenkins/plugins/main/Config.java
```diff
@@ -92,6 +94,63 @@ public void setTools(String[] tools) {
 		return this.tools;
 	}
 
+	public void setSmtpUsername(String smtpUsername) {
+		stringConfig.put("smtpUsername",smtpUsername);
```
monperrus: if the information is taken from Jenkins, we don't need those setters (duplicating info & state is always to be avoided, it is error prone)

the getters are still useful for encapsulation

src/repairnator-jenkins-plugin/src/main/java/io/jenkins/plugins/main/Config.java
```diff
@@ -7,6 +7,8 @@
 	private static Config config;
 	private File tempDir;
 	private boolean quiet = false;
+	private boolean useTLS = false;
```
monperrus: is this information in Jenkins already?

src/repairnator-jenkins-plugin/src/main/java/io/jenkins/plugins/main/RepairnatorJarDownloader.java
henry-lp: Yes, there is a bug with the Jsoup that it only parse up to 9 pushed Jar on the page and I tried many solutions on the internet, but it does not seems to help so I decided to put it away for now.

src/repairnator-jenkins-plugin/src/main/java/io/jenkins/plugins/main/Config.java
henry-lp: Unfortunately no, they only have useSSL .

monperrus: Not a problem, we don't need to do better than default Jenkins, so we can reuse their `useSSL` and remove the useTLS from here (field+g/setters+UI field)

src/repairnator-jenkins-plugin/src/main/java/io/jenkins/plugins/main/RepairnatorJarDownloader.java
monperrus: I see, what's the corresponding JSOUP bug report? Where the is JSOUP code?

src/repairnator-jenkins-plugin/src/main/java/io/jenkins/plugins/main/Config.java
henry-lp: Fixed :) .

src/repairnator-jenkins-plugin/src/main/java/io/jenkins/plugins/main/Config.java
henry-lp: done :) .

src/repairnator-jenkins-plugin/src/main/java/io/jenkins/plugins/main/RepairnatorJarDownloader.java
henry-lp: https://github.com/jhy/jsoup/issues/1259
https://stackoverflow.com/questions/21683019/jsoup-missing-content

Those are one of the solutions I tried but does not work. The most suspicious reason is that jsoup only scrape a snapshot of the page and we have to crawl for more. I would need to dig deep in the api for the correct config, so I just went with the fasted solution, just download the whole html page and parse it with Jsoup, since the repo page is usually not too big.

henry-lp:
@monperrus  
Ready for review again :)

monperrus:
Received an email from the test Jenkins, so LGTM! Will merge when CI is green (looking forward to ##1057 sigh!)

monperrus:
Thanks a lot Henry!

henry-lp:
pipeline failure due to issue (#1057) . Perhaps this can wait until I move all the docker test to the new repo :) .

henry-lp:
@monperrus  CI green. Merge ? :)

monperrus:
Great, thanks a lot!

Issue #967, opened 2019-10-22
add support for pipeline <git-url> <commit id>


monperrus:
usage:

    docker repairnator:pipeline repair-commit  <git-url> <commit id>

we assume that checkouting this commit works and then `mvn test` fails

henry-lp:
working on this.

monperrus:
great

PR #1061, opened 2020-03-29
[Refactor] refactor pipeline launcher

henry-lp:
Hi @monperrus,

I realized that we now have multiple entry points (default Travis launcher, Github, Jenkins, Listener) I decided to try refactoring for better structure. My main idea is to implement factory pattern and strategy pattern to construct the correct main process based given args input. I kept the original pipeline code and tests for comparison. 

This PR is a preview with working Github and default Travis mode. If it's good then I can also proceed refactoring Jenkins and Listener mode. WDYT :) ? 

The current Github Launcher is only another actual entry point and we need to merge them with a refactor somehow to do #967 .


monperrus:
Great idea!

- Do we plan to have two Launcher classes by keeping the original one and adding Launcher2?
- If not, what are the changes in the API of Launcher?

src/repairnator-pipeline/pom.xml
```diff
@@ -146,7 +146,7 @@
                 <configuration>
                     <archive>
                         <manifest>
-                            <mainClass>fr.inria.spirals.repairnator.pipeline.Launcher</mainClass>
```
monperrus: change to be removed, the pom.xml should not be changed in this PR.

henry-lp:
>     * Do we plan to have two Launcher classes by keeping the original one and adding Launcher2?
> 
>     * If not, what are the changes in the API of Launcher?

`branch of abstraction` - I can set the new launcher as the entry point and it will still call the old launcher by default, unless we specify in one of the option to use the new one, then once we think that everything is stable we can just remove the old launcher and the option. WDYT :) ?

monperrus:
Yes, branch by abstraction! You learned a lot in your last summer internship :-)

We first abstract from the existing class, in order to keep maximum backward compatibility. So the first refactoring would be

* Launcher implement LauncherAPI
* Launcher delegates to LegacyLauncher

> Or we can also do it in reverse . The old call the new if option is chosen for that and the pom can remain unchanged.

Yes, this is what we do.

henry-lp:
> We first abstract from the existing class, in order to keep maximum backward compatibility. So the first refactoring would be
> 
>     * Launcher implement LauncherAPI
> 
>     * Launcher delegates to LegacyLauncher

Right , I'll do that in a new PR, and rebase here after merge :) .

monperrus:
ack!

henry-lp:
@monperrus 

Here is the new entry points of the launcher :) . We have 4 different main processes (Default - Travis case, Github - repair with github, Listener - repair with ActiveMQ, Jenkins Plugin - repair as Jenkins Plugin). Each of those MainProcesses are contructed by parts - compositions as shown in graphs , each of those parts define a specific behaviour of the process (like how it define its own args for parsing and how it parse it). The contruction is done in MainProcessFactory. Thanks to this the structure is clearer and we can reuse most of the code from the Default parts and Github parts . 
RepairnatorBranchLauncher

It took time due to manual testing with Jenkins plugin due to Jenkins Plugin mode now reuse code from Github parts instead of its own after the refactor. It seems to be functioning now as before on http://jenkins.kth-assert.net , with a a problem that sometime the procedure took long time to finished (K8s Cluster cpu resources is limited ? -  I am trying to confirm it with Long ) on http://jenkins.kth-assert.net , while it works perfectly locally (my 4-core laptop).

henry-lp:
Ready for review :) . I will document the rest of the interfaces tomorrow.

monperrus:
It looks really good.

I love when you replace shared state (`Config.getInstance`) by dependency injection, it's so much better in the long term for avoiding bugs and allowing for extension.

Could you remove the space-only changes?

Since CI is green, then we can merge!

henry-lp:
@monperrus 

Yes, we wait for CI green then merge :) . I realized the problems with a singleton config when it comes to parallelization in Jenkins plugin for instance . It's better for performance (multi-threading) and separation of concerns if we use injections instead and only the Main class keep the config and distribute only necessary config info to other components, for instance we can have a factory to produce subConfigs (smaller config, its fields only dedicated to a single feature big feature) extracted from the main config after parsing the `String[] args` . As long it should by theory improve readability (now we have much fewer fields per config file), better modularity for future refactors or extension and therefore also less error-prone and easier for debugging. Also more suitable for paralellization since we have separated config now it can't cause race condition. :) .

henry-lp:
@monperrus it's green :) . Ready for review or merge .

monperrus:
Thanks Henry!

PR #1064, opened 2020-04-04
Fix: should not be null in outputdir name 

henry-lp:
should not be null in outputdir name if no commit hash provided.
Sneaky bug can cause fail in contexts such as in Jenkins.  :)


monperrus:
ci fails

henry-lp:
@monperrus It should pass now after rebase from master :) .

monperrus:
nice, tbut the diff seems wrong (far too big), the rebase/merge was likely incomplete.

henry-lp:
Strange. I fixed it , It looks correct now :) .

monperrus:
thanks, Git is always black magic :)

monperrus:
Thanks!

PR #1075, opened 2020-05-30
Feat: Jenkins Plugin update for SonarQubeRepair

henry-lp:
@monperrus 
This PR update our Jenkins Plugin to also run SonarQubeRepair .
for how the UI looks , please check (http://jenkins.kth-assert.net) in the job "repairnator-plugin-demo-pullrequest-notification" .

Result when all the rules are chosen. https://github.com/henry-lp/SonarQubeRepairTests/pull/50 .

Next: I'll read the paper you've sent me :) and perform some measurements + PR on real projects with this new tool.


src/repairnator-jenkins-plugin/src/main/java/io/jenkins/plugins/main/RepairnatorPostBuild.java
```diff
@@ -74,18 +83,44 @@
     private boolean useAstorJMut;
     private boolean useNPEFixSafe;
     private boolean useNopolTestExclusionStrategy;
+    private boolean useSonarQubeRepair;
     private final Config config = new Config();
 
+    private boolean sonarRulesGiven;
+    /* Rules */
+    private boolean rule1656;
```
monperrus, line 91: avoid boolean, prefer enum.

What about "List<SonarQubeRule>"?

src/repairnator-jenkins-plugin/src/main/java/io/jenkins/plugins/main/RepairnatorPostBuild.java
```diff
@@ -61,6 +59,17 @@
...
+import java.nio.file.Files;
+import java.nio.file.attribute.FileTime;
+import java.nio.file.attribute.BasicFileAttributes;
+
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.util.concurrent.TimeUnit;
 
 /* Post build class for post build action*/
 public class RepairnatorPostBuild extends Recorder {
```
monperrus, line 75: It's not a responsibility of RepairnatorPostBuild to handle SonarQube (or any specific repair system).

What about using the design pattern strategy?

src/repairnator-jenkins-plugin/src/main/java/io/jenkins/plugins/main/RepairnatorPostBuild.java
```diff
@@ -407,82 +585,202 @@ public Action getProjectAction(AbstractProject<?, ?> project) {
...
+        public boolean getUseNopolTestExclusionStrategy() {
             return useNopolTestExclusionStrategy;
+        }  
+
+        public boolean getUseSonarQubeRepair() {
+            return useSonarQubeRepair;
+        }
+    }
+
+    public static final class SonarRulesBlock {
```
monperrus, line 649: for sake of encapsulation, avoid public inner classes, either private inner or public top-level.

src/repairnator-jenkins-plugin/src/main/java/io/jenkins/plugins/main/RepairnatorPostBuild.java
henry-lp: This is a convention of how we should define each field for fetching into the `.jelly` (the UI file). I can make a list contains all the fields but the setter and getter need to remains otherwise your checkboxes will be unmarked in the config of Jenkins Job, despite being marked and saved before :) .

src/repairnator-jenkins-plugin/src/main/java/io/jenkins/plugins/main/RepairnatorPostBuild.java
henry-lp: Ack. I'll make this private and see if it works for the UI.

src/repairnator-jenkins-plugin/src/main/java/io/jenkins/plugins/main/RepairnatorPostBuild.java
henry-lp: A refactor will be made. :)

src/repairnator-jenkins-plugin/src/main/java/io/jenkins/plugins/main/RepairnatorPostBuild.java
henry-lp: Screenshot from 2020-06-01 21-14-37

It does not seems to be possible to change to private without getting `java.lang.IllegalAccessException`  . I can't move it either because this thing need to be a subclass to be understood as  a UI subcomponent of the main UI, otherwise the `.jelly` can't find it. 

Documentation about how to build these are very bad, I got this working mostly by piecing together the scrap info on multiple Stackoverflow pages.

src/repairnator-jenkins-plugin/src/main/java/io/jenkins/plugins/main/RepairnatorPostBuild.java
henry-lp: I just found out that this will not work with a list (I tried a HashMap). The field name need to be `rule1000` for instance then you need a setter in DescriptorImpl (the UI part) called `setRule1000` and `getRule1000` outside the descriptor . The `.jelly` infer the getter and setter name from the field name to call the correct method (example in jelly below, if we apply after specifying port it will look up getPort and setPort). Well , this means that if we have a single list as a single field then, we can't decide which to get. I tried to look for some event based stuffs to distinct the cases but the documentation of how to make a plugin is so bad. 

```
<j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
  <f:entry title="${%Port}" field="port">
    <f:textbox default="80" />
  </f:entry>
</j:jelly>
```

monperrus:
Thanks for the explanation. Ping me when you're done with the update.

henry-lp:
@monperrus  Ready for merge once Sorald is deployed :) 

Jenkins Pipeline for Sorald is configured at (http://jenkins.kth-assert.net)
Result: https://github.com/henry-lp/SonarQubeRepairTests/pull/75

monperrus:
OK thanks, is that you or @fermadeiral who would do the new SNAPSHOT deployment?

henry-lp:
@monperrus I think Fernanda :) . I have no credentials for a deployment of Sorald.

fermadeiral:
@henry-lp please, you should directly tell me when you need something from me.

fermadeiral:
First deployment with the name "sorald": https://oss.sonatype.org/content/repositories/snapshots/se/kth/castor/sorald/1.1-SNAPSHOT/

monperrus:
@henry-lp what's the status here?

henry-lp:
It's ready for merge :) .

src/repairnator-core/src/main/java/fr/inria/spirals/repairnator/GitRepositoryLauncherUtils.java
```diff
@@ -46,8 +46,7 @@ public static FlaggedOption defineArgGitRepositoryBranch() {
         FlaggedOption opt = new FlaggedOption("gitRepositoryBranch");
         opt.setLongFlag("gitrepobranch");
         opt.setStringParser(JSAP.STRING_PARSER);
-        opt.setDefault("master");
```
monperrus, line 49: it's good to have add a default value that is documented. Do we need this change?

src/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/pipeline/GitRepositoryLauncher.java
```diff
@@ -196,8 +196,6 @@ public JSAP defineArgs() throws JSAPException {
         opt = new FlaggedOption("gitbranch");
         opt.setLongFlag("gitbranch");
         opt.setStringParser(JSAP.STRING_PARSER);
-        opt.setDefault("master");
```
monperrus, line 199: we have both gitrepobranch in /GitRepositoryLauncherUtils.java and gitbranch here. That's confusing. Could we unify?

src/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/inspectors/components/RunInspector4DefaultGit.java
```diff
@@ -56,7 +55,7 @@ public void run(ProjectInspector inspector_in) {
                 }
             }
             // Add the next steps
-
+            System.out.println("skipPreSteps: " + this.skipPreSteps);
```
monperrus, line 58: replace by normal logger

src/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/step/CloneCheckoutBranchRepository.java
```diff
@@ -35,6 +35,7 @@ protected StepStatus businessExecute() {
         }
         
         String repoLocalPath = this.getInspector().getRepoLocalPath();
+
```
monperrus, line 38: avoid files with lines/space changes only

src/repairnator-core/src/main/java/fr/inria/spirals/repairnator/GitRepositoryLauncherUtils.java
henry-lp: Hi Martin, sorry for the wait :) . The reason I remove this is due to some branches does not have `master` as Default branch. I saw in `GitCheckoutBranch.java` that if the branch is null (not provided via JSAP) then the GitRepo mode will checkout the default branch automatically.

src/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/pipeline/GitRepositoryLauncher.java
henry-lp: Ah yes, this option is due to the Jenkins_Plugin mode which is now removed.  Perhaps I can remove it in the next PR as to address this issue (#1081) ? :)

monperrus:
Thanks Henry!

PR #1080, opened 2020-06-10
fix: updated repairnator-pipeline-test dependency workaround

javierron:
- Previous workaround caused repairnator-core pom files to conflict and fail CD
- It's now consistent with same type of workaround implemented in the [deploy script](https://github.com/eclipse/repairnator/blob/13da69a8d64d05d070bb5a63095f503038ac6181/.travis/travis-deploy.sh#L18)


monperrus:
Thanks!

PR #1082, opened 2020-06-12
Fix Sequencer integration

javierron:
Fixes: 
  - SequencerRepair
    - [x] Now works when used in the testPipeline test case
    - [x] No longer adds patched file to diff list
    - [x] No longer mounts the whole /tmp directory into the docker image
    - [x] Patches correctly sent to recordPatches function
    - [x] Refactor modification point search into strategy-pattern class
    - [x] Allow parallel prediction jobs
    - [x] Add `beam_size`, `image_tag`, `repairnator_threads` and `timeout` to sequencers' own config file
    - [x] Updated implementation to not start docker processes directly but instead use the DockerClient library

### DIFF
--- a/src/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/step/repair/sequencer/SequencerRepair.java
+++ b/src/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/step/repair/sequencer/SequencerRepair.java
@@ -2,22 +2,22 @@ package fr.inria.spirals.repairnator.process.step.repair.sequencer;
 
 import com.google.gson.JsonArray;
 import com.google.gson.JsonObject;
-import fr.inria.astor.approaches._3sfix.SuspiciousFile;
-import fr.inria.astor.approaches._3sfix.ZmEngine;
+import com.spotify.docker.client.DockerClient;
+import com.spotify.docker.client.messages.ContainerConfig;
+import com.spotify.docker.client.messages.ContainerCreation;
+import com.spotify.docker.client.messages.HostConfig;
+import com.spotify.docker.client.messages.Image;
+import fr.inria.spirals.repairnator.docker.DockerHelper;
+import fr.inria.spirals.repairnator.config.SequencerConfig;
 import fr.inria.spirals.repairnator.process.step.repair.sequencer.detection.ModificationPoint;
-import fr.inria.main.CommandSummary;
-import fr.inria.main.evolution.AstorMain;
 import fr.inria.spirals.repairnator.process.inspectors.JobStatus;
 import fr.inria.spirals.repairnator.process.inspectors.RepairPatch;
 import fr.inria.spirals.repairnator.process.step.StepStatus;
 import fr.inria.spirals.repairnator.process.step.repair.AbstractRepairStep;
 import fr.inria.spirals.repairnator.process.step.repair.sequencer.detection.AstorDetectionStrategy;
 import fr.inria.spirals.repairnator.process.step.repair.sequencer.detection.DetectionStrategy;
-import org.apache.commons.lang.StringUtils;
 
 import java.io.*;
-import java.lang.Runtime;
-import java.lang.Process;
 import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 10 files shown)


src/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/step/repair/sequencer/SequencerRepair.java
```diff
@@ -55,12 +57,16 @@ public String getRepairToolName() {
     protected StepStatus businessExecute() {
         this.getLogger().info("Start SequencerRepair");
         String pathPrefix = ""; // for macOS: "/private";
-        String imageTag = "repairnator/sequencer:1.0";
+        String imageTag = "javierron/sequencer:2.0";
```
monperrus, line 60: added you to dockerhub organization repairnator, you can push there instead

src/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/step/repair/sequencer/SequencerRepair.java
```diff
@@ -55,12 +57,16 @@ public String getRepairToolName() {
...
         // initJobStatus
         JobStatus jobStatus = this.getInspector().getJobStatus();
         // initPatchDir
-        this.patchDir = new File(pathPrefix + this.getInspector().getRepoLocalPath()+"/repairnator." + this.getRepairToolName().toLowerCase() + ".results");
-        this.patchDir.mkdirs();
+        Path patchDir = Paths.get(pathPrefix + this.getInspector().getRepoLocalPath()+"/repairnator." + this.getRepairToolName().toLowerCase() + ".results");
+        try {
+            Files.createDirectory(patchDir);
+        } catch (IOException e) {
+            e.printStackTrace();
```
monperrus, line 68: avoid prefer `throw new RuntimeException(e)`

src/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/step/repair/sequencer/SequencerRepair.java
```diff
@@ -202,7 +213,7 @@ protected StepStatus businessExecute() {
         this.recordToolDiagnostic(toolDiagnostic);
 
         try {
-            FileHelper.deleteFile(patchDir);
+            Files.delete(patchDir);
         } catch (IOException e) {
             e.printStackTrace();
```
monperrus, line 218: `throw new RuntimeException(e)`

monperrus:
Good progress! LGTM.

javierron:
While creating configurable options for the SequencerRepair step, I realized that maybe it is not a good idea to mix specific repair tool config with the general launcher command arguments, since it makes the launcher config even more complex than what already is, and even more, these options are not always needed.

@monperrus WDYT about a separate config file for Sequencer (maybe general to repair tools)? And maybe the path of that file can be sent through the launcher command args?

monperrus:
>  I realized that maybe it is not a good idea to mix specific repair tool config with the general launcher command arguments, since it makes the launcher config even more complex than what already is

fully agree

> WDYT about a separate config file for Sequencer

It's a good idea.

javierron:
This would be ready for now if the test pass.

However, I have a design concern:

Right now the RTScanner and FastScanner invoke the Pipeline through the RunnablePipelineContainer class, which spawns the pipeline process in a docker container. Right now the implementation of SequencerRepair spawns one docker container per patch prediction job, which leads to a docker-within-docker situation. 

This seems to be very inefficient to me, as we would have to:

- Include docker and the sequencer image within the pipeline image

or

- Pull the sequencer image for each instance of the pipeline

Both of these seem like bad ideas to me.

I will search for a solution to avoid this, maybe connecting from SeqeuncerRepair(running in the pipeline image) to the original docker service.

@monperrus is there anything I am missing or any other ideas for me to try?

javierron:
> I will search for a solution to avoid this, maybe connecting from SeqeuncerRepair(running in the pipeline image) to the original docker service.

This seems to be as easy as mounting: `-v /var/run/docker.sock:/var/run/docker.sock`

monperrus:
your analysis is 100% correct.

A third solution is to disable running the pipeline through docker, but directly as a Java process.

FTR we started using Docker for the pipeline for ensuring filesystem isolation, in particular for
Maven's `.m2`, because Github is wild. But in an enterprise setting, we can assume that the projects
under repair are well isolated in Maven and we don't need the first Docker.

monperrus:
Thanks Javier, we're making progress!

PR #1098, opened 2020-08-13
fix: fix docker pipeline runner circular reference and handle OOM error

javierron:
- RepairnatorZero: fix circular reference and handle out-of-memory error
- Renamed AlphaScanner to ZeroScanner


monperrus:
LGTM, thanks.

PR #1113, opened 2020-09-28
fix: improves Sequencer

javierron:
Astor detection strategy 
 - suspicious lines limit
 - suspicious value threshold
Sequencer collector:
 - Add context size config
Inspector:
 - Add travis build ID to "_topush" path

### DIFF
--- a/src/repairnator-core/src/main/java/fr/inria/spirals/repairnator/config/SequencerConfig.java
+++ b/src/repairnator-core/src/main/java/fr/inria/spirals/repairnator/config/SequencerConfig.java
@@ -8,12 +8,17 @@ package fr.inria.spirals.repairnator.config;
  */
 public final class SequencerConfig {
 
+    public enum RAW_URL_SOURCE {
+        RAW_GITHUB
+    }
+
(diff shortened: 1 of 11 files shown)


javierron:
I modified the functionality of the SequencerCollector to store the raw files along with the diffs.

With these files available the Sequencer training tokenizer will be able to generate training pairs extracting either the Abstract Buggy Context or the Unified Diff Context. 

Also, by doing this we keep the single-line-changes collector decoupled from any kind of context Sequencer uses to train its models.

monperrus:
Perfect. Any idea why CI fails?

javierron:
> Perfect. Any idea why CI fails?

@monperrus Yes! Because of my bad programming :sweat_smile:, but errors should be fixed now. 

Pipeline tests seem to be consistently timing out, should I split them in more jobs? Perhaps in another PR?

monperrus:
Thanks @javierron

monperrus:
> should I split them in more jobs? Perhaps in another PR?

Yes! We love, atomic, baby PR!

PR #1122, opened 2020-10-13
integrate ODS into repairnator

SophieHYe:
In this commit, we have the integrate the ODS structure into repairnator. 
Particularly, given a failing build ID, we save all generated patches in the disk, ../../patches/buildID+patchID.
Then the root path of the patches will be parsed to Coming to extract code features. 

The integration is not done yet, we will upgrade the latest coming source code in the maven repository, and the coming version in the pom file in the next commit.


src/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/inspectors/RepairPatch.java
```diff
@@ -4,135 +4,263 @@
...
+
+		} catch (PatchFailedException | IOException e) {
+			throw new RuntimeException(e);
+		}
+
+		return label;
+
+	}
+
+	private String extractODSFeatures(String rootfile) {
```
martinezmatias, line 251: Why is this method returning empty?

martinezmatias:
Hi Ye,
That's amazing!

src/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/inspectors/RepairPatch.java
```diff
@@ -4,135 +4,263 @@
...
+			File patchedFile = new File(path + "/" + buildId + "-" + patchId + "_" + buggyClassName + "_t.java");
+
+			// copy the buggy file under the patch folder
+			Files.write(Paths.get(newbuggyFile.getPath()), buggyLines);
+			// generate content of patchedFile by applying patches
+			List<String> patchedLines = DiffUtils.patch(buggyLines, patches);
+			Files.write(Paths.get(patchedFile.getPath()), patchedLines);
+
+			
+			label = extractODSFeatures(rootfile);
```
martinezmatias, line 240: the name of the method can be changed as it a) extracts features and b) executes ods

src/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/inspectors/RepairPatch.java
```diff
@@ -4,135 +4,263 @@
...
+		} catch (PatchFailedException | IOException e) {
+			throw new RuntimeException(e);
+		}
+
+		return label;
+
+	}
+
+	private String extractODSFeatures(String rootfile) {
+//		try {
```
martinezmatias, line 252: This will be the most challenging part: 
1) call coming to compute features
2) summarize the 3 set of features
3) call python

I wonder if step 2 would be directly  included/implemented in Coming

SophieHYe: Hi Matias! Fully agree with you. That's what I was thinking!  We can directly output the feature values from coming. I will make add some code on the coming repo first.

src/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/inspectors/RepairPatch.java
SophieHYe: No problem! I will change that.

SophieHYe:
> Hi Ye,
> That's amazing!
> Let's put it as Work in progress WIP, to avoid confusion

src/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/inspectors/RepairPatch.java
martinezmatias: Perfect, let me know if you need some help with any task :)

martinezmatias:
I change the status of this PR to "Draft"

src/repairnator-pipeline/src/test/java/fr/inria/spirals/repairnator/process/inspectors/TestGlobalPatchAnalysis.java
```diff
@@ -41,156 +43,162 @@
...
+		List<RepairPatch> classifyPatcheswithODS = inspector.getJobStatus().getCorrectbessLabeledPatches(Features.ODS,
+				buildId);
+
+		// There are 10 patches are generated for this failing build
+		assertThat(classifyPatcheswithODS.size(), is(10));
+
+		int overfittingCount = 0;
+
+		for(RepairPatch patch : classifyPatcheswithODS) {
+			if ("overfitting".equals(patch.getODSLabel())){
```
martinezmatias, line 194: [minor] We can use an enumerator with all the possible values that ODS can generate. 
Actually, we can create one enum from this tree fields: https://github.com/SpoonLabs/coming/blob/ef0451b97086d803bf954c539b726ca6e0f82d4e/src/main/java/fr/inria/coming/codefeatures/RepairnatorFeatures.java#L36

SophieHYe: Hi Matias,

I thought about the enumerator before, then I realized it may be less necessary for that. As in our code, we did not compare any type of ods labels in the logic, we simply return the string type label according to probability score. WDYT?

SophieHYe: Hi Matias, I thought I leave comments but not succeeded :) 
I actually thought about add an enumerator of the ODS labels, then after analysis, I think it may not be necessary.  In our code, we do not have the logic comparison or other usage of the labels, we simply return the string value according to the probability score, and this value will be set as a property of the repair patch. Creating an enumerator, we will have still eventually return a string. WDYT?

martinezmatias (commented):
> we simply return the string value according to the probability score, and this value will be set as a property of the repair patch. Creating an enumerator, we will have still eventually return a string. WDYT?

We can return the enumerator's values e.g.

```
enum PATCH_LABEL{
   OVERFITTING, CORRECT, UNKNOWN, ....;
}
```

then the method returns a type `PATCH_LABEL`  such as (Pseudo-code):

```
public  PATCH_LABEL getLabel(patch){
    
        if(ODS.isOverfitting(patch)) 
         return  PATCH_LABEL.OVERFITTING;
        else
       return  PATCH_LABEL.CORRECT;

}

```

this would avoids us to then compared the output via strings (which can be error proud -typos in the string, case sensitive, etc-)

src/repairnator-pipeline/src/test/java/fr/inria/spirals/repairnator/process/inspectors/TestGlobalPatchAnalysis.java
martinezmatias: > we simply return the string value according to the probability score, and this value will be set as a property of the repair patch. Creating an enumerator, we will have still eventually return a string. WDYT?

We can return the enumerator's values e.g.

```
enum PATCH_LABEL{
   OVERFITTING, CORRECT, UNKNOWN, ....;
}
```

then the method returns a type `PATCH_LABEL`  such as (Pseudo-code):

```
public  PATCH_LABEL getLabel(patch){
    
        if(ODS.isOverfitting(patch)) 
         return  PATCH_LABEL.OVERFITTING;
        else
       return  PATCH_LABEL.CORRECT;

}

```

this would avoids us to then compared the output via strings (which can be error proud -typos in the string, case sensitive, etc-)

SophieHYe:
@martinezmatias The latest coming version is updated. Please review. Thanks!

martinezmatias:
Hi @SophieHYe 
great, thanks! I'll review it.

src/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/inspectors/JobStatus.java
```diff
@@ -291,7 +291,14 @@ public void addPatches(String toolName, List<RepairPatch> patches) {
         }
         return allPatches;
     }
-
+    
+    
+    protected List<RepairPatch> getCorrectbessLabeledPatches(Features features, Long buildId) {
```
martinezmatias, line 296: Do you mean `getCorrectnessLabeledPatches`  (i.e., getCorrect**n**essLabeledPatches) instead of getCorrect**b**essLabeledPatches?

martinezmatias:
Hi @SophieHYe 
The PR seems almost perfect! Just one typo in one method name (see comment just before). After fixing that, we are ready for merging!
Thanks a lot for the nice work!

SophieHYe:
@martinezmatias, 
The typo is corrected in the new commit, please review :)

martinezmatias:
Great, thanks!

martinezmatias:
merged, thanks @SophieHYe nice work

monperrus:
Indeed, big kudos @SophieHYe

PR #1123, opened 2020-10-13
Refactor: move items in pipelineb folder into pipeline

henry-lp:


monperrus:
Thanks!

martinezmatias:
The commit https://github.com/eclipse/repairnator/pull/1123/commits/c7208f16ae87eeca5344bfc848105b6471145fb2 renames the file (due to package rename) and modifies the file to reflect the package change.
However, the commits does not rename those files from the test: https://github.com/eclipse/repairnator/tree/master/src/repairnator-pipeline/src/test/java/fr/inria/spirals/repairnator/pipelineb
Thus there are compilation error.

martinezmatias:
(I can do the refactoring but I am not sure what is/was the purpose of such "*b" package)

henry-lp:
Hi @martinezmatias :), it is because we have 3 different modes Travis, Github and ActiveMQlistener, which are more or less being tightly coupled to each other in `LegacyLauncher.java` back then. Initially,  `LegacyLauncher.java` was only used as the entry point when Repairnator is used  Travis. Later, we added ActiveMQListener mode and  Github mode where both are more or less branched from the original Travis mode (a.k.a. LegacyLauncher) and all mode switching happens in the LegacyLauncher.  ActiveMQlistener mode is currently only build on Travis meaning that Repairnator in this mode only repairs based on Travis build id. The problem is for instance if a future implementation which require ActiveMQListener to repair for instance a Github would be very likely to introduce more technical debt. Therefore, I back then in this [PR](https://github.com/eclipse/repairnator/pull/1061) proposed that we break down the three modes into smaller components to employ dependency injections instead. Moreover, this was regarded as experimental back then therefore we did this with "branch by abstraction" to still keep the original while gradually go over to the new one or revert completely to the

martinezmatias:
Hi @henry-lp  

Thanks a lot for the clarification!
Regards
Matias

Issue #1057, opened 2020-03-23
Continuous integration pipeline is flaky


monperrus:
As you all know, we have long-standing problems with the Repairnator CI:
- it's too slow
- it's flaky

Having look at the build time without tests (only dependencies & compile), the Docker pull jobs take ages.

So I propose to move all tests requiring docker to a separate repo for integration testing, it could be:
- a Travis job: https://github.com/repairnator/repairnator-integration-tests
- a Jenkins job: https://ci.eclipse.org/repairnator/job/repairnator-integration-test

WDYT?

henry-lp:
I am on this :) .

monperrus:
great!

henry-lp:
Hi @monperrus 

I check a bit and only RTScanner have a docker test. 

Waiting for merging this PR (https://github.com/repairnator/repairnator-integration-tests/pull/2)

Also , it would be great if I have access to the repo https://github.com/repairnator/repairnator-integration-tests , so I can set up a CI pipeline :) .

henry-lp:
Also, would you also want to move these test cases (they are unrelated to docker, but they are a bit heavy) ?

https://github.com/eclipse/repairnator/blob/17749319ba290695058b260aaf1063777c3938c9/src/repairnator-realtime/src/test/java/fr/inria/spirals/repairnator/realtime/TestRTScanner.java#L45-L103

henry-lp:
Also, I just check and they seems to not have `Docker` installed on their slave pods on Eclipse Jenkins, because docker tests use [this dependency](https://github.com/spotify/docker-client) to build DockerRunnablePipeline and they require `Docker 1.6+` to run .

monperrus:
great progress!

> Waiting for merging this PR (repairnator/repairnator-integration-tests#2)

merged, thanks

> Also , it would be great if I have access to the repo repairnator/repairnator-integration-tests 

added you as admin: https://github.com/repairnator/repairnator-integration-tests/invitations

> would you also want to move these test cases (they are unrelated to docker, 

not now (baby PR always :-)

> Perhaps we should go with a Travis job

yes, let's do this for now in https://github.com/repairnator/repairnator-integration-tests/

monperrus:
Reopening this one, CI is again very flaky, more mocking is necessary. We can analyze the last 50 builds to identify the flaky tests to refactor.

monperrus:
FYI
Root causing, detecting, and fixing flaky tests: State of the art and future roadmap
[...]

PR #1134, opened 2020-10-30
fix: update R-Hero pipeline with ODS

javierron:
Here is a working integration of ODS into the R-Hero pipeline.

**Other fixes**
- @SophieHYe @martinezmatias, I changed the directory where ODS saves the buggy-patched file pairs. The path used before was the one provided by the repair tool, and those paths are temporary, and get deleted after the pipeline finishes (at least on SequencerRepair). Any thoughts? Also, I think that this path should be configurable.

- Updated the feedback links on the PR text as URL encoding was not a good idea after all.

- Updated the PR text and title to be +9000 fancy. :tophat:

**Discussion**
Right now this works, but I don't like that I have called the ODS function from within the repair step. Perhaps later we can move overfitting detection to it's own pipeline step? or create a sort of switch in the AbstractRepairStep class, so that one could write in the pipeline definition for example:

```
AbstractRepairStep step = new RepairStepXYZ().SetODSLabels(true)
```

@monperrus WDYT?

### DIFF
--- a/src/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/inspectors/RepairPatch.java
+++ b/src/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/inspectors/RepairPatch.java
@@ -189,32 +189,29 @@ public class RepairPatch {
 
 		try {
 			// create a directory to store the patch: "patches/"+buildId+patchId
-			String buggyClassName = buggyFile.getName().split(".java")[0];
-			String rootfile = filePath.split("patch")[0] + "/patches";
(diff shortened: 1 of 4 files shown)


monperrus:
Thanks Javier, LGTM. CI seems to fail because of flakiness #1057
 
Agree with your design problem about the ODS integration, but it's low priority for now. Could you create an issue to remember it?

OK to merge?

javierron:
Tests run locally, I would say it's safe to merge.

Opened #1135

monperrus:
Thanks @javierron

PR #1136, opened 2020-11-02
chore: make Repairnator compile on Java >= 9

### LINKED ISSUE #1133: Make Repairnator compile on Java >= 9
Nopol requires tools.jar as a dependency.
This library was replaced in [JEP 220](https://openjdk.java.net/jeps/220).

When we try to compile Repairnator on a JDK which does not have tools.jar we get `BUILD FAILED`. Example, with OpenJDK 14:
`
ERROR] Failed to execute goal on project repairnator-pipeline: Could not resolve dependencies for project fr.inria.repairnator:repairnator-pipeline:jar:3.3-SNAPSHOT: Could not find artifact com.sun:tools:jar:1.4.2 at specified path /usr/lib/jvm/java-14-openjdk/../lib/tools.jar -> [Help 1]
`

Ideally, we should make Repairnator compile and run (except for Nopol) on any JDK version that does not have tools.jar (any OpenJDK or Oracle Java >= 9)

andre15silva:
Closes #1133


src/maven-repair/src/main/java/com/github/tdurieux/repair/maven/plugin/NopolMojo.java
```diff
@@ -55,6 +56,16 @@
...
 	@Override
     public void execute() throws MojoExecutionException, MojoFailureException {
+        try {
+            // Check if tools.jar is in classpath, as Nopol depends on it.
+            // This is not done in compile time as we don't want Repairnator to it to compile on OpenJDK or JDK >= 9
+            URLClassLoader loader;
+            loader = (URLClassLoader) ClassLoader.getSystemClassLoader();
+            loader.loadClass("com.sun.jdi.AbsentInformationException");
+        } catch (ClassNotFoundException e) {
+            getLog().error("tools.jar has not been loaded, therefore Nopol can't run");
```
monperrus, line 66: throw new RuntimeException(e) instead of logging

src/repairnator-core/src/main/java/fr/inria/spirals/repairnator/states/PipelineState.java
```diff
@@ -21,5 +21,6 @@
     PLUGINSNOTCOMPUTED,
     PROPERTY_FILE_NOT_WRITTEN,
     BUG_FAILING_PASSING,
-    BUG_PASSING_PASSING
+    BUG_PASSING_PASSING,
+    MISSING_DEPENDENCIES
```
monperrus, line 25: nice addition

src/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/step/repair/nopol/AbstractNopolRepair.java
```diff
@@ -86,6 +88,14 @@ public void initWithJobStatus() {
         this.setSources(jobStatus.getRepairSourceDir());
     }
 
+    public void checkToolsJar() throws ClassNotFoundException {
+    	// Check if tools.jar is in classpath, as Nopol depends on it.
+        // This is not done in compile time as we don't want Repairnator to it to compile on OpenJDK or JDK >= 9
+        URLClassLoader loader;
```
monperrus, line 94: duplicate code with NopolMojo? Reuse the same method?

monperrus (commented):
Promising first contribution. Here are a couple of things to change. Thanks André!

src/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/step/repair/nopol/AbstractNopolRepair.java
andre15silva: I didn't do this initially because they are in different modules. Should I move the checkToolsJar() method to repairnator-core then?

monperrus:
yes, let's do that!

andre15silva:
I moved the method to `repairnator-core`.

To use it in `maven-repair`, I had to add `repairnator-core` as a dependency. 
In `repairnator-pipeline`, the version of the dependency is given by the property `revision` in `repairnator-root`, but in this project this property is not accessible, as the parent project is not `repairnator-root`.

What version of `repairnator-core` should `maven-repair` use? Ideally it should come from `repairnator-root`.

monperrus:
> To use it in `maven-repair`, I had to add `repairnator-core` as a dependency. 

That's good.

> What version of `repairnator-core` should `maven-repair` use? Ideally it should come from `repairnator-root`.

Yes.

Your work here is very interesting. In Maven-Repair we encapsulate repair tools per Mojo. We do the same in the pipeline. For example, NopolMojo is more or less the same as NopolRepairStep. This duplication is costly for maintenance. It would be good to unify this, for example NopolMojo may depend on and call NopolRepairStep. WDYT?

monperrus:
Sidenote: could you do the paper work on the Eclipse website so that check eclipsefdn/eca  is green? Thanks.

andre15silva:
> > What version of `repairnator-core` should `maven-repair` use? Ideally it should come from `repairnator-root`.
> 
> Yes.

I'll work on that and making CI pass.
 
> Your work here is very interesting. In Maven-Repair we encapsulate repair tools per Mojo. We do the same in the pipeline. For example, NopolMojo is more or less the same as NopolRepairStep. This duplication is costly for maintenance. It would be good to unify this, for example NopolMojo may depend on and call NopolRepairStep. WDYT?

Sounds good to me. Created #1137 to keep track of it. I can tackle it after this one.

> Sidenote: could you do the paper work on the Eclipse website so that check eclipsefdn/eca is green? Thanks.

Done!

monperrus:
> I'll work on that and making CI pass.

OK. Note that we have flaky tests in some jobs.

andre15silva:
Ready for review @monperrus .

Two small notes:

- I had to create two profiles, one for having Nopol (which works when the JDK includes `tools.jar`) and another for all other Java versions. They have automatic activation according to the Java version. For OpenJDK 8 it's necessary to specify has the version is 1.8 but does not include tools.jar and Maven only supports prefix matching automatic activation.
- This PR doesn't make Repairnator work for NPEFix on Java >= 9 version, as `maven-repair` is used by NPEFix's step on `repairnator-pipeline`. See https://github.com/eclipse/repairnator/issues/1137#issuecomment-723585969.

monperrus:
Thanks a lot for this excellent first contribution @andre15silva

PR #1153, opened 2020-11-26
Refactor: Unify encapsulation of NpeFix

monperrus:
```
/home/andre/Repos/npe-dataset/lang-304/target/test-classes/:/home/andre/Repos/npe-dataset/lang-304/target/classes/:/home/andre/.m2/repository/junit/junit/4.7/junit-4.7.jar:/home/andre/Repos/npe-dataset/lang-304/target/npefix/npefix-bin/:/home/andre/.m2/repository/fr/inria/gforge/spirals/npefix/0.7/npefix-0.7.jar
/home/andre/Repos/npe-dataset/lang-304/target/npefix
org.json.JSONException: JSONObject["executions"] not found.Since we have surefire plugin configure to true, to automatically trigger NPEFixWe don't break the build for errors during the build, continue.
```

PR #1163, opened 2020-12-11
feat: add localMavenRepository config parameter

andre15silva:
This is a parameter for `repairnator-pipeline`, which will be used in #1153 , as NPEFix needs this for building the classpath.

Separate PR just to avoid making #1153 too cluttered.


andre15silva:
Also, I'm not sure if the Jenkins plugin needs this parameter.

monperrus:
Thanks a lot, it's a good idea.

It's not elegant to have to modify four classes to do this:

* TravisDefineJSAPArgs.java
* GithubDefineJSAPArgs.java
* LegacyLauncher.java
* GitRepositoryLauncher.java

What about doing a refactoring first to define the command line args at a single location in the code?

andre15silva:
Is there any reason why only the GitHub mode has options for Sorald?

monperrus:
> Is there any reason why only the GitHub mode has options for Sorald?

No. We will align everything as much as possible.

Are you finished here?

andre15silva:
> > Is there any reason why only the GitHub mode has options for Sorald?
> 
> No. We will align everything as much as possible.

Roger that.


> Are you finished here?

No, it's still WIP. But almost finished.

andre15silva:
It's pretty much ready. There's only one part left to unify, which is the repair tools argument and initialization.

The definition is not possible to unify because it depends on `RepairToolsManager` for the `StringParser`. 
For this we can either have a `PipelineLauncherUtils` class or have these `Launcher`s be child of an `AbstractLauncher`. 
The problem with the latter approach is that `Travis` and `Github` are `MainProcess` and not `Launcher`s.

Another problem is that not all `Launcher`s have the same `LauncherMode` initialization. Just to double check, we also want this unified right?

monperrus:
> not all Launchers have the same LauncherMode initialization.

What do you mean by "initialization" in this context?

> There's only one part left to unify, which is the repair tools argument and initialization.

Per the practice of doing "baby pull-requests", the changes here are good enough. We can merge here in order to resume and finish #1164 (highest priority) and #1153

andre15silva:
> What do you mean by "initialization" in this context?

The initialization of `RepairnatorConfig`.
Example:
https://github.com/eclipse/repairnator/blob/989ee8375a2b0020fd8058ce9ec0abc23aaba376/src/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/pipeline/LegacyLauncher.java#L395
https://github.com/eclipse/repairnator/blob/989ee8375a2b0020fd8058ce9ec0abc23aaba376/src/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/pipeline/travis/TravisInitConfig.java#L128

> 
> > There's only one part left to unify, which is the repair tools argument and initialization.
> 
> Per the practice of doing "baby pull-requests", the changes here are good enough. We can merge here in order to resume and finish #1164 (highest priority) and #1153

src/repairnator-core/src/main/java/fr/inria/spirals/repairnator/dockerpool/BuildAnalyzerLauncher.java
```diff
@@ -6,7 +6,7 @@
 import com.martiansoftware.jsap.JSAPResult;
 import fr.inria.spirals.repairnator.InputBuildId;
 import fr.inria.spirals.repairnator.LauncherType;
-import fr.inria.spirals.repairnator.LauncherUtils;
```
monperrus, line 9: We avoid moving classes, it easily breaks client code. Could you move it back? Thanks!

andre15silva: Done!

monperrus:
Thanks!

PR #1164, opened 2020-12-14
Enable tests in CI

andre15silva:


monperrus:
That's an essential PR! :+1::+1::+1:

andre15silva:
Is there anything needed to be activated in Jenkins so that changes in the `Jenkinsfile` take effect?

I ran the `repairnator-pipeline` manually on a docker container created with the updated image, and there should be no failures and 10 errors. On Jenkins it produced the same result as before (3 failures and 11 errors).

Jenkins is pulling the new image from what I saw in the logs tho, so I'm a bit confused as to why this happens.

andre15silva:
Pipeline tests are passing when I run the stage on a local docker container, but are failing in our CI. I'm investigating why this happens.

andre15silva:
@javierron, is there any specific reason why ODSPath is, by default, under the home directory? One of the failing tests is failing because of that, maybe we could move it to the output folder.

javierron:
@andre15silva You can move it where it's needed, as long as it is not deleted after the pipeline is executed.

andre15silva:
I think I might be useful to have this plugin https://github.com/jenkinsci/junit-plugin to be able to access the test report, as the errors only happen in Jenkins.

andre15silva:
A stage (in https://ci.eclipse.org/repairnator/job/repairnator-ci/job/PR-1164/19/) is running in an infinite loop. Can someone with permission cancel it @javierron @monperrus ?

andre15silva:
> I think it might be useful to have this plugin [jenkinsci/junit-plugin](https://github.com/jenkinsci/junit-plugin) to be able to access the test report, as the errors only happen in Jenkins.

@javierron could you take a look at this too, while we're at it?

PR #1165, opened 2020-12-18
fix: fix jsap arg handling regression

andre15silva:
There was a mistake in the last PR (#1163) which I, stupidly, didn't notice. 

This PR fixes it.


monperrus:
Thanks. This is meant to be catch by a test.

Is there a test case specifying this behavior? (assuming CI works again or that we run the tests locally)

If not, could you add one?

andre15silva:
> Is there a test case specifying this behavior? (assuming CI works again or that we run the tests locally)

There wasn't any for this case specifically. The behavior should appear in other tests, but because they are failing I didn't notice.

> If not, could you add one?

I added a test that should cover similar behavior in the default Launcher. I could had tests for the other types of Launchers, but I think that fixing the tests should be prioritary.

monperrus:
Thanks for the fix!

>  Since the tests are not working, I think that's probably prioritary right?

100% yes!

PR #1166, opened 2021-01-04
Re-enable tests in CI (Eclipse Jenkins)

javierron:
I'm hijacking @andre15silva's PR #1164 to attempt some fixes myself. Any progress will be rebased to his PR.


andre15silva:
Just to keep you up to date then, the current problems are:

1. The `pipeline` stage works on the local docker image, but on the Jenkins CI it gives those errors.
2. `maven-repair` has a problem, when running locally in the docker container, with `spoon-core` jar file not being available in the local repository. I'm trying to reproduce that in Jenkins.
3. Yesterday `maven-repair` hit an infinite loop. I'm not sure what is causing it. I'm adding a timeout so that doesn't happen again.

javierron:
> Just to keep you up to date then, the current problems are:
> 
>     1. The `pipeline` stage works on the local docker image, but on the Jenkins CI it gives those errors.
> 
>     2. `maven-repair` has a problem, when running locally in the docker container, with `spoon-core` jar file not being available in the local repository. I'm trying to reproduce that in Jenkins.
> 
>     3. Yesterday `maven-repair` hit an infinite loop. I'm not sure what is causing it. I'm adding a timeout so that doesn't happen again.

Cool! thanks for the info! I am focusing on point 3 for now. I believe the problem might be here:
https://github.com/eclipse/repairnator/blob/fc73f4a0c506a961a610a2e9d271fc91f1b8bb63/src/maven-repair/src/test/java/com/github/tdurieux/repair/maven/plugin/NPEfixMojoTest.java#L14

My guess is that the `mvn clean test` command is failing and the tests waits forever. I'll keep you updated.

andre15silva:
Could be. When running on the local docker container the only problem is with the Nopol test not finding the spoon-core jar, so should be related to the environment in Jenkins.

javierron:
Solved 2 & 3 [link](https://ci.eclipse.org/repairnator/job/repairnator-ci/view/change-requests/job/PR-1166/22/console). I will attempt point 1 tomorrow and hopefully the modification on the Jenkinsfile will fix some of the problems of the pipeline.  @andre15silva

javierron:
This PR is passing the tests now.
@andre15silva are there any new changes on your branch? If not, we can skip the rebase and merge this PR instead.

Jenkinsfile
```diff
@@ -12,69 +12,103 @@ spec:
...
+    }
+    stage('repairnator-maven-repair'){
+      environment {
+        TEST_PATH="src/maven-repair/"
+        TEST_LIST="**"
+      }
+      steps {
+        container('maven') {
+          sh 'bash ./.ci/ci-maven-repair.sh'
         }
```
andre15silva, line 93: Is there really a need to use a separate script for `maven-repair`? It would be nice to merge this with `run-with-core`. The only difference is the `NPEFix` version option.

If we change this we can eliminate `xmlstarlet` from the Docker image too.

src/maven-repair/pom.xml
```diff
@@ -146,6 +146,21 @@
...
+        <dependency>
+            <groupId>commons-cli</groupId>
+            <artifactId>commons-cli</artifactId>
+            <version>1.3</version>
+        </dependency>
+        <dependency>
+            <groupId>org.json</groupId>
+            <artifactId>json</artifactId>
+            <version>20160810</version>
+        </dependency>
```
andre15silva, line 163: I'm not sure this is the best approach, albeit it works. Shouldn't the dependencies be fetched transitively?

andre15silva:
> This PR is passing the tests now.
> @andre15silva are there any new changes on your branch? If not, we can skip the rebase and merge this PR instead.

The only changes were the timeouts, but you have already included that here. I added some comments, tell me what you think. 

Other than that, we can skip the rebase and merge this yes.

src/maven-repair/pom.xml
javierron: > I'm not sure this is the best approach

I agree with you. At least in the case of `spoon-core` I could see that the dependency metadata (pom and sha files) is downloaded, but unless I put these here explicitly it does not download the jar. I assumed the same for `commons-cli` and `json`.

Also, I checked and the maven-repair tests work locally in my pc only because those jars were already there as a result of compiling another module.

If there is a better way to trigger the jar download, we can replace this.

Jenkinsfile
javierron: I'm not really sure of what does that version property does, have you tried to run the tests without it?

andre15silva: Yes, and they work. I also don't know why that is needed. I haven't found any usage of a property `NPEFIX_VERSION`.

javierron: @cokefh @monperrus WDYT?

src/maven-repair/pom.xml
andre15silva: Yes, it's very weird behavior indeed. While debugging I also saw that pom and sha files were downloaded, just not the jar file.

I think it's important to get this right tho, because having sub-dependencies hardcoded makes it hard to maintain. 

I will look further into it after my exam tomorrow.

andre15silva: `mvn dependency:tree` for `maven-repair` is in https://gist.github.com/andre15silva/254c0de8ef2d2a7ac45c834f404119c4

The issue is that these three dependencies are being omitted because of conflicts with sub-dependencies from other repair tools that use these libraries but with a different version.

```
[INFO]    +- (fr.inria.gforge.spoon:spoon-core:jar:7.0.0:compile - omitted for conflict with 6.0.0)
[INFO]    +- (commons-cli:commons-cli:jar:1.3:compile - omitted for conflict with 1.2)
[INFO]    \- (org.json:json:jar:20160810:compile - omitted for conflict with 20160212)
```

If we added the `Nopol` dependency first, it would fetch these versions. It's also not a very elegant solution tho. Yours is probably better, as it is explicitly written.

Another option would be to change `getClassPathFromPom` to take into account that the version explicited in the repair tools pom files may not be available locally and that another version has to be used. I'm not sure this is a very good idea tho.

monperrus:
Thanks a lot for the heavy work. LGTM to merge and proceed. We can fine-tune later.

OK for you?

src/maven-repair/pom.xml
javierron: > If we added the Nopol dependency first, it would fetch these versions. It's also not a very elegant solution tho. Yours is probably better, as it is explicitly written.

I believe I tried this, and it did not resolve the conflicts.

> Another option would be to change getClassPathFromPom to take into account that the version explicited in the repair tools pom files may not be available locally and that another version has to be used.

I don't think this is a good idea either, we may create an even more complicated compatibility/dependency issue.

javierron:
> OK for you?

OK for me

andre15silva:
> > OK for you?
> 
> OK for me

Ok for me too

monperrus:
Thanks a lot for the foundational PR

PR #1171, opened 2021-01-18
chore: repairnator-core for NPEFix unification

andre15silva:
- Adds default value for localMavenRepository parameter
- Move `Pair` from `maven-repair` to `repairnator-core`. Original class is still included. We need this one to avoid cyclic dependencies between modules.

We need to decide if we keep the original `Pair` class. It is not used anywhere other than NPEFix.

### DIFF
--- a/src/repairnator-core/src/main/java/fr/inria/spirals/repairnator/config/RepairnatorConfig.java
+++ b/src/repairnator-core/src/main/java/fr/inria/spirals/repairnator/config/RepairnatorConfig.java
@@ -705,6 +705,9 @@ public class RepairnatorConfig {
     }
 
     public String getLocalMavenRepository() {
+        if (localMavenRepository == null) {
+            setLocalMavenRepository(System.getenv("HOME") + "/.m2/repository");
+        }
         return localMavenRepository;
     }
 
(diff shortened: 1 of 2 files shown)


monperrus:
Thanks. Instead of re-inventing `Pair` it would be better to reuse that of Apache Commons Lang, see https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/tuple/Pair.html

andre15silva:
> Thanks. Instead of re-inventing `Pair` it would be better to reuse that of Apache Commons Lang, see [commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/tuple/Pair.html](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/tuple/Pair.html)

Yup, good idea. Should we remove the re-invented `Pair` from `maven-repair` too then? It was already there, hence why I didn't remove it.

monperrus:
> Should we remove the re-invented Pair from maven-repair too then?

Yes, that would be a useful PR

andre15silva:
> Yes, that would be a useful PR

Ok. The unification removes all usages of it. So we can maybe remove it after the unification.

monperrus:
OK for me.

andre15silva:
It's ready. I say we replace `Pair` after all these PRs.

monperrus:
Thanks @andre15silva we're making good progress.

PR #1174, opened 2021-01-22
chore: fix junit CI plugin config

andre15silva:
Moves junit plugin command to a post action. Fixes commands not being run if stage fails.


andre15silva:
This is ready, CI is running again because of a connection that timed out during a test.

PR #1175, opened 2021-01-22
feat: add addNextSteps method

andre15silva:
- Implements addNextSteps method
- Implements test for addNextSteps method
- Implements test for addNextStep method


monperrus:
Thanks @andre15silva for the baby PR

PR #1177, opened 2021-01-24
feat: introduce class RunInspector4Maven

andre15silva:
- Introduces RunInspector4Maven
- Test for RunInspector4Maven

### DIFF
--- a/src/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/inspectors/InspectorFactory.java
+++ b/src/repairnator-pipeline/src/main/java/fr/inria/spirals/repairnator/process/inspectors/InspectorFactory.java
@@ -42,9 +42,9 @@ public class InspectorFactory {
 		return inspector;
 	}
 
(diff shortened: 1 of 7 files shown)


andre15silva:
Had a little (unrelated) hitch. Should be ready.

Next one needs this one merged first.

monperrus:
We're almost there:

- The rebase with master is not complete, we should not see addNextSteps in the diff
- Could you add some API Javadoc in RunInspector4Maven?

Thanks!

andre15silva:
>     * The rebase with master is not complete, we should not see addNextSteps in the diff
Fixed.

>     * Could you add some API Javadoc in RunInspector4Maven?
Added. Could you check to see if it is ok?

monperrus:
Good, thanks. I'll merge after CI.

andre15silva:
Might be a good idea to abort the ongoing job and the one queued after that, to save some time, as we don't have parallel jobs yet.

monperrus:
Thanks @andre15silva

Issue #1186, opened 2021-02-05
Bump Java version to Java 11


andre15silva:
Related #371.

Need to be careful with `Nopol`.

andre15silva:
@monperrus I can jump straight on this, as this is blocking compilation on `master`

andre15silva:
This is actually trickier than I thought. It breaks more than Nopol switching to Java 11, particularly tests. Needs further analysis.

monperrus:
I'm not surprised that a few tests need to be adapted.

monperrus:
Per our conversation with @javierron, on Astor, we may have problems with GZoltar, which means we would revive @martinezmatias 's essential long-term work on Flacoco, see #1116

PR #1187, opened 2021-02-08
Downgrade Sorald's version so that Repairnator compiles again on Java 8

andre15silva:
Sorald no longer supports Java 8 (see https://github.com/SpoonLabs/sorald/issues/371

This change is meant to be temporary and be reverted after we drop support for Java 8.

### DIFF
--- a/src/repairnator-pipeline/pom.xml
+++ b/src/repairnator-pipeline/pom.xml
@@ -37,7 +37,7 @@
         <dependency>
             <groupId>se.kth.castor</groupId>
             <artifactId>sorald</artifactId>
-            <version>1.1-SNAPSHOT</version>
+            <version>1.1-20200928.114859-2</version>
         </dependency>
         <dependency>
             <groupId>fr.inria.repairnator</groupId>
@@ -124,7 +124,6 @@
             <artifactId>activemq-core</artifactId>
             <version>5.7.0</version>
         </dependency>
-        
     </dependencies>
 
     <properties>


monperrus:
We can do that temporarily, esp since @khaes-kth is working on his fork.

However, CI is red.

andre15silva:
Yes, Java VM is crashing during `TestPipelinebGithubMode`. I'm trying to fix it.

khaes-kth:
I think @javierron had found a workaround for this problem. However, I am fine with downgrading for now. Later we will probably need to change it back to the latest version.

andre15silva:
> I think @javierron had found a workaround for this problem. However, I am fine with downgrading for now. Later we will probably need to change it back to the latest version.

As of know, we can compile it. The problem would be with the tests, particularly Nopol, and I was also getting some issues with the junit plugin (probably just needs an upgrade).


> Yes, Java VM is crashing during `TestPipelinebGithubMode`. I'm trying to fix it.

I changed Sorald's version. The one I was using was making the VM run OOM, I think it should be green after it runs everything now.

javierron:
> I think @javierron had found a workaround for this problem. 



> As of know, we can compile it.

Yes, I managed to compile it without problem, but also had to revert because Astor + GZoltar were not behaving correctly with java 11.

monperrus:
Thanks a lot @andre15silva this is not anymore a blocker for the other parallel PRs.

Yet, we all agree that #1186 is the highest priority task now. WDYT?

andre15silva:
> Yet, we all agree that #1186 is the highest priority task now. WDYT?

Sounds good. I'll work on that straight away, while merging NPEFix refactor too.

monperrus:
deployed 17f106c8e7b3aea25542966410c0b26c2e6c2e43-SNAPSHOT/ corresponding to 17f106c8e7b3aea25542966410c0b26c2e6c2e43 as a workaround for the java 8 problem in Repairnator

monperrus:
FTR, the workaround seems successful: https://github.com/eclipse/repairnator/pull/1192

PR #1205, opened 2021-03-29
refactor: SequencerCollectorScanner uses GithubActions instead of TravisCI

javierron:
* Changed SequencerCollectorScanner's source to GitHub actions.
* Modified @khaes-kth GitHub scanner to also provide an "only passing" filter.


monperrus:
Thanks, it's an important change. Could we add a test case of this new feature?

javierron:
@monperrus test case added for the new only-passing filter of the Github scanner.

monperrus:
Thanks @javierron

Issue #1176, opened 2021-01-22
Flaky tests


andre15silva:
There are some flaky tests currently.

I have identified two so far:

- https://github.com/eclipse/repairnator/blob/103e6415f012a58b4686a92c66b8d761f0f96253/src/repairnator-pipeline/src/test/java/fr/inria/spirals/repairnator/pipeline/TestPipeline.java#L55	[Failed](https://ci.eclipse.org/repairnator/job/repairnator-ci/job/PR-1174/1/) and [Passed](https://ci.eclipse.org/repairnator/job/repairnator-ci/job/PR-1174/2/)	
Reason: `org.eclipse.jgit.api.errors.TransportException: Connection timed out (Connection timed out)`

- https://github.com/eclipse/repairnator/blob/103e6415f012a58b4686a92c66b8d761f0f96253/src/repairnator-pipeline/src/test/java/fr/inria/spirals/repairnator/pipelineb/TestPipelinebGithubMode.java#L67	[Failed](https://ci.eclipse.org/repairnator/job/repairnator-ci/job/PR-1169/1/) and [Passed](https://ci.eclipse.org/repairnator/job/repairnator-ci/job/PR-1169/3/)
Can't determine reason because there's no access to test report

monperrus:
Can we close this one?

andre15silva:
This still happens from time to time, so I think it's better to keep it open.

andre15silva:
Another flaky test identified:
- https://github.com/eclipse/repairnator/blob/103e6415f012a58b4686a92c66b8d761f0f96253/src/repairnator-pipeline/src/test/java/fr/inria/spirals/repairnator/process/step/TestGatherTestInformation.java#L109
[Failed](https://ci.eclipse.org/repairnator/job/repairnator-ci/job/PR-1204/2/testReport/junit/fr.inria.spirals.repairnator.process.step/TestGatherTestInformation/testGatherTestInformationOnlyOneErroring/) and [Passed](https://ci.eclipse.org/repairnator/job/repairnator-ci/job/PR-1204/4/)
Reason: sometimes the error message of the test project returns two lines without `\n` separating them

monperrus:
TestMetrics4BearsJsonFile.testBearsJsonFileWithPassingPassingBuilds is flaky, last instance is https://ci.eclipse.org/repairnator/job/repairnator-ci/job/monperrus-patch-2/1/console

PR #1208, opened 2021-04-06
refactor: Overfitting detection as an AbstractRepairStep component

### LINKED ISSUE #1135: Overfitting detection as pipeline step or AbstractRepairStep component
Right now the ODS function is called from within the SequencerRepair repair step. We should refactor this and move overfitting detection to it's own pipeline step, or create a sort of switch in the AbstractRepairStep class, so that one could write in the pipeline definition, for example:

```
AbstractRepairStep step = new RepairStepXYZ().SetODSLabels(true)
```

andre15silva:
Closes #1135

> Why not a separate pipeline step?

Currently repair steps call `recordPatches` when they finish their work, which calls the notifiers. If, after that, we have logic that can filter/rank/etc the patches, it messes up that logic.

> What is there to finish?

Are the patch ranking methods for ODS working (tests are ignored). If so, wouldn't it make more sense to have different modes for the different features? Instead of `OVERFITTING` we would have `ODS_P4J` and `ODS_S4R` for example. @monperrus wdyt?

Also, should ranking be done inside each repair step or overall? The later makes more sense to me, but would require a separate step for it and would mess up the observer pattern mentioned above.

## ChangeLog
- Add missing parameters to control patch ranking, filtering and classification to `LauncherUtils` and `RepairnatorConfig`. Update test accordingly.
- Move ranking (to finish), filtering and classification logic to a method in `AbstractRepairStep`. Update test accordingly. Ranking and filtering tests to be activated/written.

### DIFF
--- a/src/repairnator-core/src/main/java/fr/inria/spirals/repairnator/LauncherUtils.java
+++ b/src/repairnator-core/src/main/java/fr/inria/spirals/repairnator/LauncherUtils.java
@@ -132,11 +132,6 @@ public class LauncherUtils {
         // --sequencerRepair
         jsap.registerParameter(LauncherUtils.defineArgSequencerRepairMode());
 
-        // --patchRanking
-        jsap.registerParameter(LauncherUtils.defineArgPatchRanking());
-        // --patchRankingMode
-        jsap.registerParameter(LauncherUtils.defineArgPatchRankingMode());
-
         // --patchClassification
         jsap.registerParameter(LauncherUtils.defineArgPatchClassification());
         // --patchClassificationMode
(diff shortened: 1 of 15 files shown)


monperrus:
Cool to make progress on this topic.

> Are the patch ranking methods for ODS working (tests are ignored).

We can remove the ranking code (outdated plus not researched anymore). We'll start from scratch again with #1210 

> If so, wouldn't it make more sense to have different modes for the different features

No. We want to simply take the best feature set identified by @SophieHYe in the [ODS paper](http://arxiv.org/pdf/1910.12057).

andre15silva:
This is ready. The only test failing is a flaky test (identified in #1176 )

monperrus:
Thanks. What's the intention behind PATCH_FILTERING_MODE?

monperrus:
> The only test failing is a flaky tes

Could you remove / deflakify it in a separate PR? Thanks!

andre15silva:
> Thanks. What's the intention behind PATCH_FILTERING_MODE?

In the case there is a new filtering mode in the future (in the case there are more than one filtering options for the same classification mode), the developer will just need to add a new mode.

> Could you remove / deflakify it in a separate PR? Thanks!

Done. Opened #1214

andre15silva:
Ready

monperrus:
Great, thanks.

PR #1217, opened 2021-05-05
doc: add bibliographic info abour r-hero

monperrus:


javierron:
Oh, what a nice article :D


Test fails because of flaky test (as reported on #1176)

PR #1232, opened 2021-06-18
Bump revision to 3.5-SNAPSHOT

andre15silva:
We have released up to revision `3.4` on `repairnator-core`.

### DIFF
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -69,7 +69,7 @@
     </repositories>
 
     <properties>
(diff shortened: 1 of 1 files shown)


andre15silva:
Description and url had to be added to `repairnator-root`'s `pom.xml` due to sanity checks by Maven Central during https://github.com/eclipse/repairnator/pull/1201#issuecomment-865045338

PR #1216, opened 2021-04-26
feat: add exception parser

monperrus:
Move class from old repository https://github.com/Spirals-Team/exception-parser to here.

We need that to finish https://github.com/eclipse/repairnator/pull/1201

PR #1235, opened 2021-08-09
chore: remove non-existent package

monperrus:
this code is now in core, see #1216


monperrus:
flaky test, restart CI

PR #1255, opened 2022-08-29
connectAnonymously removed, now PatchFilterTest uses Jenkins env variables

Sofi1410:


monperrus:
looks great, let's keep the PR open!

